Issue #752, opened 2024-07-04
Remove the `key` field from `KeyedTypeABI::Tuple` and `KeyedTypeABI::Map`


mohammadfawaz:
They are not necessary. Only primitive types should get keys... All other keys can be inferred from them.

mitchmindtree:
@mohammadfawaz are arrays of tuples also flattened? E.g. 

```
storage {
    foo: { int, int }[3],
}
```
for each element, does this result in the following keys:

- [0]
- [1]
- [2]
- [3]
- [4]
- [5]

or these:

- [0, 0]
- [0, 1]
- [1, 0]
- [1, 1]
- [2, 0]
- [2, 1]

And is the behaviour the same for a tuple of arrays? E.g. `{ int[2], int[2], int[2] }`.

just checking as this does complicate abi-gen key construction a bit further still, but might be doable... will report back.

mohammadfawaz:
Yup! Arrays of tuples are flattened all the way.

mitchmindtree:
> Yup! Arrays of tuples are flattened all the way.

And just double-checking, is the case the same for tuples containing arrays?

mohammadfawaz:
> > Yup! Arrays of tuples are flattened all the way.
> 
> 
> 
> And just double-checking, is the case the same for tuples containing arrays?

Absolutely. We think of everything as a pile of data, except for maps, and likely other dynamic containers in the future.

mitchmindtree:
Closed via #779.

PR #767, opened 2024-07-09
refactor(pint-abi-gen): Add `pint-abi-visit`, programmatic key construction

### LINKED ISSUE #762: Change `pint-abi-gen` to construct keys for storage/pub-vars programmatically
Opening a dedicated issue for the approach originally described here: https://github.com/essential-contributions/pint/issues/752#issuecomment-2212460367.

> we could potentially remove the key from the ABI output altogether if we standardise and document how storage type nesting maps to key construction. E.g.
>
> - Top-level storage "fields" are prefixed by a single word field index.
> - Array elements are prefixed by a single word array index.
> - Tuple elements are prefixed by a single word tuple index.
> - Map entries are prefixed by the map key encoded as words.

> This might also make it a bit easier to support dynamically sized keys (e.g. strings or byte arrays) in the future, as the current key representation in the ABI output (`Vec<Option<Key>>` with `None`s for map keys) can't support unknown sized keys.

@mohammadfawaz pointed out that we will likely want to continue flattening tuples and arrays. While these complicate the rules slighlty, it should still be doable as long as the necessary context required for flattening is provided during traversal.

## Steps

- [x] Move the `pint-abi-gen` refactor commit from #761 with the `visit` mod into its own PR.
    - Done: #767.
- [x] Move the `visit` module into its own `pint-abi-visit` crate that can be shared between `pint-abi` and `pint-abi-gen` for handling traversals of the `KeyedTypeABI` tree.
- [x]  Change `KeyElem` to `Nesting` in `pint-abi-visit` crate.
- [x]  In `Nesting` type, change `Fixed` to `Var` for storage/pub-key var, and add a `TupleIx` variant.
- [x]  Change the key stack in generated `Mutations` to be `Vec` of type `enum KeyElem { MapKey(Vec<Key>), ArrayIx(Word) }` (declared in `pint-abi`).
- [x]  Update `pint_abi::merge_key` to merge `[KeyElem]` into a `[Nesting]`.
    - **This is where key flattening can happen** - e.g. consecutive `Nesting::ArrayIx`s can be collapsed into a single `Word`.
    - This still requires some thought around how to derive the flattened key from nested tuples. The `Nesting::TupleField` will likely need both `index` and `flattened_index` fields...
- [x]  Change `abi_key_expr`s to be `nesting_expr`s with type `[Nesting]` rather than `[Option<Word>]` for use with the updated `merge_key` function.

These changes will also help to separate the concept of keys (which can sometimes be flattened) from the concept of type nesting which is required for human-readable API generation that matches the user's storage/pub var declarations.

mitchmindtree:
This is a pretty major refactor of the `pint-abi-gen` crate ahead of landing support for arrays, `keys()` builders, and more.

## `pint-abi-visit`

The major addition is the `pint-abi-visit` crate which provides some helpers for traversing the `[KeyedVarABI]`s of `storage` and `pub vars`. For context, `pint-abi-gen` does a traversal of the nested ABI types in order to generate types and impls for array, map and tuple types that allow for building mutations in a type-safe manner.

When visiting each `KeyedTypeABI`, `struct` and `impl` generation require a lot more context than just the current type being visited. E.g. in order to be able to produce a unique struct name, provide docstrings, and produce a key construction expression, we need to know the full "`Nesting`" of the type. In order to be able to produce and `impl` with builder methods for nested types, we also need to be able to iterate over the nested "child" nodes within the currently visited type.

To assist with all of this, we now first flatten the `[KeyedVarABI]`s into an indexable tree structure. Then rather than recursively traversing through the `KeyedTypeABI`s, we instead traverse the tree using indices. The indexed approach gives us the freedom to traverse the parent or child nodes as necessary when generating items, which wasn't really possible with the recursive `KeyedTypeABI` traversal.

## Key Construction

This PR also changes the way that keys are constructed for mutations. **We no longer use the keys specified in the ABI JSON**, and instead use the current context of the `[KeyedVarABI]` traversal to determine the keys programmatically. cc @mohammadfawaz 

Currently, we just use the keys from the leaf types (e.g. `Bool`, `Int`, etc) to verify that the key construction matches the compiler's key construction, but this can probably be removed along with the keys from the ABI in general. This does increase the risk of running into a discrepancy between the ABI-gen key construction and the `pintc` key construction slightly, but hopefully these cases should be caught by the `pint-abi-gen-tests` before they can be released.

Ideally we would share the approach to key construction between `pintc` and `pint-abi-gen` somehow, but I haven't dug into the compiler enough to know how feasible this is.

Closes #762.
Related #752.
Dependency of #761.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1185,6 +1185,7 @@ name = "pint-abi-visit"
 version = "0.0.1"
 dependencies = [
  "essential-types",
+ "petgraph",
  "pint-abi-types",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1140,6 +1140,7 @@ dependencies = [
  "essential-types",
  "pint-abi-gen",
  "pint-abi-types",
+ "pint-abi-visit",
  "serde_json",
  "thiserror",
 ]
@@ -1150,6 +1151,7 @@ version = "0.1.0"
 dependencies = [
  "essential-types",
  "pint-abi-types",
+ "pint-abi-visit",
  "proc-macro2",
  "quote",
  "serde_json",
@@ -1176,6 +1178,14 @@ name = "pint-abi-types"
 version = "0.1.0"
 dependencies = [
  "serde",
+]
+
+[[package]]
+name = "pint-abi-visit"
+version = "0.0.1"
+dependencies = [
+ "essential-types",
+ "pint-abi-types",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ members = [
   "pint-abi-gen",
   "pint-abi-gen-tests",
   "pint-abi-types",
+  "pint-abi-visit",
   "pint-cli",
   "pint-manifest",
   "pint-pkg",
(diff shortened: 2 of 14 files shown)


mohammadfawaz (approved):
Amazing! Thanks for this.. I see that it was quite painful to deal with flattened arrays and tuples.

With this, we now need to officially and formally standardize key assignment in a spec. I'll open an issue about that.

Issue #771, opened 2024-07-09
ABI gen not handling anon namespace


freesig:
![image](https://github.com/essential-contributions/pint/assets/7286454/a4e47de3-857f-40e6-9c87-7ab0a4429f26)
This PR shows the issue https://github.com/essential-contributions/essential-integration/pull/49

This is the abi:
```json
{
  "predicates": [
    {
      "name": "::Burn",
      "vars": [
        {
          "name": "__::I_pathway",
          "ty": "Int"
        },
        {
          "name": "::anon@0::sig",
          "ty": {
            "Tuple": [
              {
                "name": null,
                "ty": "B256"
              },
              {
                "name": null,
                "ty": "B256"
              },
              {
                "name": null,
                "ty": "Int"
              }
            ]
          }
        },
        {
          "name": "::anon@0::public_key",
          "ty": {
            "Tuple": [
              {
                "name": null,
                "ty": "B256"
              },
              {
                "name": null,
                "ty": "Int"
              }
            ]
          }
        }
      ],
      "pub_vars": [
        {
          "name": "::token",
          "ty": {
            "Tuple": [
              {
                "name": "contract",
                "ty": {
                  "B256": [
                    0,
                    0
                  ]
                }
              },
              {
                "name": "addr",
                "ty": {
                  "B256": [
                    0,
                    1
                  ]
                }
              }
            ]
          }
        }
      ]
    },
    {
      "name": "::Cancel",
      "vars": [
        {
          "name": "__::I_pathway",
          "ty": "Int"
        },
        {
          "name": "::anon@1::sig",
          "ty": {
            "Tuple": [
              {
                "name": null,
                "ty": "B256"
              },
              {
                "name": null,
                "ty": "B256"
              },
              {
                "name": null,
                "ty": "Int"
              }
            ]
          }
        },
        {
          "name": "::anon@1::public_key",
          "ty": {
            "Tuple": [
              {
                "name": null,
                "ty": "B256"
              },
              {
                "name": null,
                "ty": "Int"
              }
            ]
          }
        }
      ],
      "pub_vars": [
        {
          "name": "::token",
          "ty": {
            "Tuple": [
              {
                "name": "contract",
                "ty": {
                  "B256": [
                    0,
                    0
                  ]
                }
              },
              {
                "name": "addr",
                "ty": {
                  "B256": [
                    0,
                    1
                  ]
                }
              }
            ]
          }
        }
      ]
    },
    {
      "name": "::Mint",
[...]

PR #772, opened 2024-07-09
fix: handle macro declared vars

freesig:
If you declare a var inside a macro it gets a name like `::anon@0::v5`. This handles that by also replacing `@` and non-prefix `::` with `_`.

Maybe don't merge until @mitchmindtree has checked this

### DIFF
--- a/pint-abi-gen/src/lib.rs
+++ b/pint-abi-gen/src/lib.rs
@@ -710,8 +710,8 @@ fn mutation_method_from_keyed_var(name: &str, ty: &KeyedTypeABI) -> syn::ImplIte
 fn mutations_methods_from_keyed_vars(vars: &[KeyedVarABI]) -> Vec<syn::ImplItemFn> {
     vars.iter()
         .map(|var| {
-            let name = strip_colons_prefix(&var.name);
-            mutation_method_from_keyed_var(name, &var.ty)
+            let name = field_name_from_var_name(&var.name);
+            mutation_method_from_keyed_var(&name, &var.ty)
         })
         .collect()
 }
--- a/pint-abi-gen/src/lib.rs
+++ b/pint-abi-gen/src/lib.rs
@@ -73,16 +73,16 @@ fn ty_from_pint_ty(ty: &TypeABI) -> syn::Type {
 /// Names for fields are emitted by pint with a `::` prefix.
 /// This function checks for the `::` prefix and strips it if it exists.
 fn strip_colons_prefix(name: &str) -> &str {
-    let mut split = name.split("::");
-    let first = split.next();
-    split.next().or(first).expect("name had unexpected format")
+    name.trim_start_matches("::")
 }
 
-/// Var names have the `::` prefix, and sometimes have `.` separators for
-/// flattened tuple fields. We strip the `::` prefix, and replace all `.`
+/// Var names have the `::` prefix, and sometimes have `.` or `@` or `::` separators for
+/// flattened tuple fields. We strip the `::` prefix, and replace all `.` or `@` or `::`
 /// occurrences with `_`.
 fn field_name_from_var_name(name: &str) -> String {
-    strip_colons_prefix(name).replace('.', "_")
+    strip_colons_prefix(name)
+        .replace(['.', '@'], "_")
+        .replace("::", "_")
 }
 
 /// A named field for each of the decision variables.
--- a/pint-abi-gen-tests/test-pkgs/simple/src/contract.pnt
+++ b/pint-abi-gen-tests/test-pkgs/simple/src/contract.pnt
@@ -23,6 +23,8 @@ predicate Foo {
     pub var t1: int;
     pub var t2: b256;
 
+    @bar();
+
     // Here we write a bunch of arbitrary constraints to check encoding/decoding
     // of the different types etc.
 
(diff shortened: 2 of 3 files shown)


freesig:
fixes #771

freesig:
Actually this doesn't handle pub vars declared in macros. Fixing this now

pint-abi-gen/src/lib.rs
```diff
@@ -73,16 +73,16 @@ fn ty_from_pint_ty(ty: &TypeABI) -> syn::Type {
...
 
-/// Var names have the `::` prefix, and sometimes have `.` separators for
-/// flattened tuple fields. We strip the `::` prefix, and replace all `.`
+/// Var names have the `::` prefix, and sometimes have `.` or `@` or `::` separators for
+/// flattened tuple fields. We strip the `::` prefix, and replace all `.` or `@` or `::`
 /// occurrences with `_`.
 fn field_name_from_var_name(name: &str) -> String {
-    strip_colons_prefix(name).replace('.', "_")
+    strip_colons_prefix(name)
+        .replace(['.', '@'], "_")
```
otrho, line 84: My only concern here is the reason we used `@` in the first place.  By replacing it with `_` you could introduce a name clash.

freesig: Like if someone names a var `anon_v5`?

otrho: Yep.

freesig: We might just have to take the risk. It will only affect abi gen. Although it would suck if you are trying to use the abi of an already deployed contract that has this name collision. I mean you could just change the name because that won't change the address of the contract.

I'm not sure what else we can do because rust only allows `[a-zA-Z][a-ZA-Z0-9_]*` for vairable / fields / functions.

otrho: Yeah, I'm just thinking using Rust identifiers for Pint identifiers might not be the best.  An API that uses strings for identifiers instead of actual Rust identifiers would be safer.

freesig: It makes it impossible to use Rust types then which is kind of the whole point of this.

If it's just strings then you are back to guessing what the fields are. Basically you need to keep looking at the json manually

mitchmindtree (approved):
I'm kind of surprised we can have macros that expand to var/pub-var declarations, but looks like this PR just helps support existing stuff so :shipit:

freesig:
>I'm kind of surprised we can have macros that expand to var/pub-var declarations, but looks like this PR just helps support existing stuff so :shipit:

If we couldn't do this then it greatly reduces the ability to reuse code with macros. The vars/ pub vars are still scoped to the macro so they can't be used outside the macro.

pint-abi-gen/src/lib.rs
freesig: @mitchmindtree before I merge this what are your thoughts on this?

mitchmindtree:
I guess it just feels a bit odd that we abstract them away in a macro in pint, but then they're revealed on the application side in ABI-gen anyways.

My intuition is that like state, var and pub var declarations should be explicit as they're all the inputs to the predicate. Having them hidden in macros kind of feels like hiding function arguments :sweat_smile:

I'm struggling to think of cases where you wouldn't want to declare a `var`/`pub var` with some type, and then pass the var ident as an input to the macro instead to do whatever other constraint expansion etc is necessary :thinking: Maybe the use-case will become clearer to me once I actually start hacking on some application stuff.

freesig:
>I'm struggling to think of cases where you wouldn't want to declare a var/pub var with some type, and then pass the var ident as an input to the macro instead to do whatever other constraint expansion etc is necessary 🤔 Maybe the use-case will become clearer to me once I actually start hacking on some application stuff.

Basically something like this:
https://github.com/essential-contributions/essential-integration/blob/d226ba333d1a27488589f298181a4deb84fe9961/apps/token/pint/signed/src/contract.pnt#L109

pint-abi-gen/src/lib.rs
mitchmindtree: I'm not too concerned with this as the ABI-gen is just a thing to improve devex, and if a name conflict does happen the Rust compiler will just complain about an already declared method or field.

freesig:
I guess if we decide to not allow declaring vars and pub vars in macros then I could not merge this PR and just change the integration apps to not do that?

@mitchmindtree another thing that this kind of fixed but might not actually be a problem is that currently if you have a name like `::foo::bar` it becomes `foo` not `foo_bar`. I'm not sure if it's possible to declare a var or pub var in another file but if it is this would be a problem.

freesig:
I might just merge this as it's blocking work on the integration repo. If we decide to make it not possible to declare var or pub var in macros then we can revert this change later.

mitchmindtree:
> I'm not sure if it's possible to declare a var or pub var in another file

I don't think it is, and I suspect we don't want to allow this. It might be odd to declare an input for a predicate somewhere outside of the predicate itself? Again, this might just be a self-imposed mental block I have by imagining vars and pub vars as function inputs for predicates, but it's the closest mental model I can think of. I feel like as someone reading the code, you want to be able to read all the inputs in one place, but this might just be personal preference.

I'll open a separate issue to discuss decision variable / transient data decls in macros :+1:

PR #779, opened 2024-07-14
Remove keys from the JSON ABI

mohammadfawaz:


mohammadfawaz:
@mitchmindtree this PR is not 100% ready.. I still need to update the docs. But do you mind checking if this is all sane?

otrho (approved):
LGTM.

mitchmindtree (approved):
Yep this LGTM!

> I still need to update the docs

Yeah having some docs explaining how keys can be derived from the type nesting would be :ok_hand:

PR #780, opened 2024-07-18
WIP* Allow `Map` to be declared outside of `Storage Block`

### LINKED ISSUE #574: Allow `Map` to be declared outside of `Storage Block`
Currently the `parser` only allows `map` declarations inside of `storage blocks`. We want to allow a `map` to be declared earlier then used in the storage block.

Ex.
```pnt
type myMap = (b256 => int);

results in

Error: expected `::`, `b256_ty`, `bool_ty`, `ident`, `int_ty`, `real_ty`, `string_ty`, or `{`, found `(`
    ╭─[test.pnt:17:14]
    │
 17 │ type myMap = (b256 => int);
    │              ┬  
    │              ╰── expected `::`, `b256_ty`, `bool_ty`, `ident`, `int_ty`, `real_ty`, `string_ty`, or `{`
────╯

----


Vs. a storage block has no issues

storage {
    balances: (b256 => int),
}

```

lethalgem:
Closes #574 

Allows:
```pnt
declaring map at the global level to be used in a storage block

type MyMap = (int => int);
type MyInt = int;
type MyNestMap = (MyInt => int);

storage {
    m: MyMap,
    n: MyInt,
    o: MyNestMap,
}
```

Disallows:
```pnt
using map type declaration as a var

predicate test {
    type MyMap = (int => int);

    var y: MyMap;
    var x: ( int => int );
}

and declaring a map type inside of the storage block

storage {
    type MyMap = (b256 => int);
}
```

Let me know if there are other test scenarios I should include/work on


pintc/tests/storage/new_type_map.pnt
```diff
@@ -0,0 +1,31 @@
...
+// type ::MyMap = ( int => int );
+// type ::MyInt = int;
+// type ::MyNestMap = ( ::MyInt => int );
+// >>>
+
+// flattened <<<
+// storage {
+//     m: ( int => int ),
+//     n: int,
+//     o: ( ::MyInt (int) => int ),
```
otrho, line 26: This doesn't seem to have lowered the inner alias.  This should happen in `transform::lower_aliases()` in `lower.rs`.

When glancing at the code in there I see it has a recursive inner function for finding alias types and replacing them with the aliased type, but it doesn't go further when the replace type is also an alias.

So here it's replacing `MyNestMap` with `(MyInt => int)` but it doesn't bother continuing to replace `MyInt` with `int`.  I have a feeling this is a problem in general, not just this PR.  As in an alias of an alias probably won't be lowered properly for `var` or `state` types either.

The solution I think is to iterate until there are no changes/replacements made, in the same fashion as other transformation passes, since the replacing type could itself be quite complex and have many nested type aliases.

And so this is a tricky enough problem to warrant its own issue.  I'll do that now and approve this PR.

otrho (approved):

otrho (approved):

Issue #774, opened 2024-07-10
Refactor the `Contract`/`Predicate` data types.


otrho:
While implementing the compiler the design of the language has changed, from having a single 'root' implicit predicate only, to having a 'root' predicate along with optional 'child' predicates, to now having no 'root' predicate and only explicit named predicates.

The data structures used to represent a collection of predicates is now unfit for use.  In particular, much of what was in the implicit predicate is copied into the child predicates as they're created, which is data that can go stale and also breaks values which are in a global namespace, especially `const` declarations.

The solution is to move most of what is in a `Predicate` back into the `Contract`.  The predicate declarations have namespaces though and some data need to remain in the explicit predicate structs.

But which items should be allowed to be declared within `predicate { ... }` blocks?  Currently the parser accepts anything except macros, storage, interfaces and consts.

This is probably OK.  We could perhaps also disallow enum and new-type declarations (or they should at least be stored in the `Contract` and namespaced to the `Predicate`).  Vars, state, constraints, if/else, non-expression macro calls and instances are the other items currently allowed.

All expressions must be stored in the `Contract` and only referred to from the `Predicate`s though.

### Tasks

* [x] Move `exprs` from `Predicate` to `Contract`.
* [x] Use a `slotmap` for `preds` and use a `PredKey` to refer to predicates instead of a reference or by name.
* [x] Streamline `Predicate`:
    * [x] Move storage from `Predicate` to `Contract`.
    * [x] Move `enums` from `Predicate` to `Contract`.
    * [x] Move `new_types` from `Predicate` to `Contract`.
    * [x] Move `interfaces` from `Predicate` to `Contract`.
* [x] Remove the concept of a 'root' predicate.
* [x] Re-enable `const` arrays.
* [x] Fix constant resolution in type aliases to arrays.  (See E2E test `root_types`.)
* [x] Refactor `WithPred` to either be `WithCtx` which includes the `Contract` or make it `WithContract` which takes a `PredKey` argument.

PR #783, opened 2024-07-19
Move `exprs` from `Predicate` into `Contract`.

otrho:
This is the first task in #774.  It doesn't add any new functionality, is just a refactor to put `exprs` in `Contract`.

It's a bit of a mess and the rest of the tasks in #774 will clean it up.

The type checker and the transforms are changed the most.  Anywhere we used to need a `Predicate` reference we now pretty much also need a `Contract` reference.  For the type checker, which likes to update types in-place as they're resolved, and for the transforms, which are obviously making wholesale changes to the predicates, this is a bit of a borrowing nightmare.  Consequently a lot of the time instead of passing `pred: &Predicate`s around we're now passing `pred_name: &str`s and the predicate is fetched on demand from the contract by name.

This isn't ideal and so the next task is to put the `Predicate`s into a `slotmap` and refer to them generally by `PredKey` instead of `&Predicate` or by name.

But this PR is a change which touches a lot of the code and rebasing/merging against it is already a pain.  (If #780 is merged first then this will get conflicts, but then vice versa.  It might be easier to merge this one first, as the smaller PR is easier to reason about -- I can fix the conflicts.)

### DIFF
--- a/pintc/src/asm_gen.rs
+++ b/pintc/src/asm_gen.rs
@@ -46,7 +46,9 @@ pub fn compile_contract(
 
     for (name, pred) in contract.preds.iter() {
         if name != Contract::ROOT_PRED_NAME {
-            if let Ok(predicate) = handler.scope(|handler| predicate_to_asm(handler, pred)) {
+            if let Ok(predicate) =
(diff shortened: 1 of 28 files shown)


mohammadfawaz (approved):
👍👍

PR #787, opened 2024-07-22
feat(pint-abi): Generate `ADDRESS` consts for contracts, predicates

mitchmindtree:
This adds support for generating `ADDRESS` constants for contracts and predicates by optionally providing the path to the compiled contract associated with the ABI.

The `pint_abi::gen_from_file!` macro API has been changed slightly. Previously, it was used like so:

```rust
pint_abi::gen_from_file!("path/to/abi.json");
```

Now, the invocation looks like:

```rust
pint_abi::gen_from_file! {
    abi: "path/to/abi.json",
    contract: "path/to/contract.json",
}
```

The `contract` field is optional. When provided, the contract at the specified path will be used to determine the contract and predicate `ADDRESS` consts and include them in their associated generated modules. If not provided, the old behaviour remains the same.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1148,11 +1148,13 @@ dependencies = [
 name = "pint-abi-gen"
 version = "0.1.0"
 dependencies = [
+ "essential-hash",
  "essential-types",
  "pint-abi-types",
  "pint-abi-visit",
  "proc-macro2",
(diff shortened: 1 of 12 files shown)


mohammadfawaz (approved):
Cool!

PR #790, opened 2024-07-23
Restrict `enum`s and `type`s to be only at the top level.

otrho:
And per #774 move them from `Predicate` to `Contract`.  And move `storage` too, but that's fairly trivial.


mohammadfawaz (approved):
Awesome :shipit:

PR #792, opened 2024-07-24
feat: Add cargo publish action

freesig:
Adds in an action that will publish crates that have been bumped when a PR is merged.


mohammadfawaz (approved):
Awesome, thanks!

PR #793, opened 2024-07-24
Only generate `.json` on successful compile

### LINKED ISSUE #704: Only generate `.json` output on successful compile.
The current `master` will drop the ABI JSON file even if the compile fails, which might be fine but feels weird to me.

lethalgem:
Closes #704 

Will no longer generate a `-abi.json` unless the contract can be compiled. This means any compilation error during `parsing`, `flattening`, and `asm-gen` will prevent the file from being generated/modified

ex.
```pnt
--parse error--
predicate test {
    type MyMap = (int => int);
}

--flatten error--
predicate test {
    var x = 2;
    var y: ( int => int );
}

--asm-gen error--
predicate test {
    state x = __storage_get(123);
}

### DIFF
--- a/pintc/src/main.rs
+++ b/pintc/src/main.rs
@@ -47,36 +47,37 @@ fn main() -> anyhow::Result<()> {
         }
     };
 
-    // Produce `json` ABI
-    let mut filepath_stem = filepath
-        .file_stem()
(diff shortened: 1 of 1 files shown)


mohammadfawaz (approved):
Great! Thanks.

PR #794, opened 2024-07-24
Lower Nested `Alias`

### LINKED ISSUE #782: Nested aliases aren't lowered correctly.
This doesn't seem to have lowered the inner alias.  This should happen in `transform::lower_aliases()` in `lower.rs`.

When glancing at the code in there I see it has a recursive inner function for finding alias types and replacing them with the aliased type, but it doesn't go further when the replace type is also an alias.

So here it's replacing `MyNestMap` with `(MyInt => int)` but it doesn't bother continuing to replace `MyInt` with `int`.  I have a feeling this is a problem in general, not just this PR.  As in an alias of an alias probably won't be lowered properly for `var` or `state` types either.

The solution I think is to iterate until there are no changes/replacements made, in the same fashion as other transformation passes, since the replacing type could itself be quite complex and have many nested type aliases.

_Originally posted by @otrho in https://github.com/essential-contributions/pint/pull/780#discussion_r1683661569_

lethalgem:
Closes #782

### DIFF
--- a/pintc/tests/basic_tests/aliases.pnt
+++ b/pintc/tests/basic_tests/aliases.pnt
@@ -1,0 +1,70 @@
+type MyInt = int;
+type MyReal = real;
+type MyBool = bool;
+type MyTuple = { int, int };
+type MyArray = bool[5];
+type MyMap = (b256 => int);
+type MyB256 = b256;
+type MyNestedAlias = MyInt;
+
+storage {
+    r: MyMap,
+}
+
+predicate test {
+   var m: MyInt;
+   var n: MyReal;
+   var o: MyBool;
+   var p: MyTuple;
+   var q: MyArray;
+   var s: MyB256;
+   var x: MyNestedAlias;
+}
+
+// parsed <<<
+// type ::MyInt = int;
+// type ::MyReal = real;
+// type ::MyBool = bool;
+// type ::MyTuple = {int, int};
+// type ::MyArray = bool[5];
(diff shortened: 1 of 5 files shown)


pintc/tests/basic_tests/aliases.pnt
```diff
@@ -0,0 +1,70 @@
+type MyInt = int;
+type MyReal = real;
+type MyBool = bool;
+type MyTuple = { int, int };
+type MyArray = bool[5];
+type MyMap = (b256 => int);
```
mohammadfawaz, line 6: I'd also like to see a map type here that uses two aliases as in:
```
type MyMap = ( MyB256 => MyTuple )
```

Also, maybe some quite complex such as 
```
type MyOtherMap = ( MyArray => MyMap )
```

lethalgem: `MyOtherMap` isn't allowed right now. The keys for a map must be int, bool, or b256.

```pnt
type MyOtherMap = ( MyArray => MyMap );

results in

Error: compiler internal error: currently in storage maps, keys must be int, bool, or b256 and values must be int or bool
    ╭─[test.pnt:23:5]
    │
 23 │     z: MyComplexMap,
    │     ───────┬───────  
    │            ╰───────── currently in storage maps, keys must be int, bool, or b256 and values must be int or bool
```

mohammadfawaz (approved):
Nice, thanks!

PR #799, opened 2024-07-26
Some more `Contract` / `Predicate` refactoring.

otrho:
As per #774 this PR takes us a couple more steps towards not having a 'root' predicate.

- `interfaces` are now in `Contract`.
- `top_level_symbols` are now just `symbols` and have been wrapped in a struct.  The plan is to have an instance in `Contract` for all actual top-level symbols, and each predicate would have their own `symbols` which may not shadow the top-levels, but won't clash with other predicates.  TBC.
- `Expr::PathByKey` is gone, now we just have `Expr::PathByName` which is renamed to `Expr::Path`.  This removes the dependency between `Expr` (a globally used struct) and `VarKey` (only significant in predicates).  As a consequence there is now also no dependency between `Type` and predicates.
- Add `DisplayWithContract` which is the same as `DisplayWithPred` except just for contract contexts (like displaying `Expr` and `Type`).  This has simplified a bunch of code which needed a predicate just to generate a `Display` string, especially places like the type checker and its error messages, and has reduced the need for using `Contract::root_pred()` just as a stand-in for `DisplayWithPred`.

We're not quite there but I think pretty close to removing the 'root' predicate which will simplify things further.

### DIFF
--- a/pintc/src/asm_gen.rs
+++ b/pintc/src/asm_gen.rs
@@ -368,7 +368,7 @@ impl AsmBuilder {
                 s_asm.extend(asm.iter().map(|op| StateRead::Constraint(*op)));
 
                 // Get the `interface` declaration that the storage access refers to
-                let interface = &pred
+                let interface = &contract
                     .interfaces
                     .iter()
                     .find(|e| e.name.to_string() == *interface_instance.interface)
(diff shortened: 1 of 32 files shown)


mohammadfawaz (approved):
Nice nice
image

PR #800, opened 2024-07-26
Add `intrinsic` names to the `lexer`

### LINKED ISSUE #580: Add intrinsic names to the lexer.
After adding intrinsic support we're adopting leading underscores to differentiate their names.

Currently the parser will parse intrinsic calls with `"ident"` tokens and then accepts or rejects them based on leading `__`.

A preferred method, if possible, would be for the lexer to differentiate between regular identifers and intrinsic names, relieving the parser from needing to do the checks manually.

lethalgem:
Closes #580

### DIFF
--- a/pintc/src/error/parse_error.rs
+++ b/pintc/src/error/parse_error.rs
@@ -26,8 +26,6 @@ pub enum ParseError {
     },
     #[error("expected identifier, found keyword `{keyword}`")]
     KeywordAsIdent { span: Span, keyword: String },
-    #[error("expected identifier, found intrinsic name `{name}`")]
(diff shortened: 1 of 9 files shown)


otrho (approved):
Logos uses this strange method for disambiguating matches.  Most lexers will just choose the first one in the list if there are multiple matches (as is the case here with `Ident` and `IntrinsicName`) but Logos chooses the most specific.  And having the `__` at the front is slightly more specific. 🙂 

The 'was expecting ident but got `__foo`' error message is a bit annoying, perhaps one day if it comes up we can change the error to 'but got intrinsic name'.

PR #802, opened 2024-07-29
Remove the 'root' predicate.

otrho:
Again as per #774 this is the next step in refactoring `Predicate`.  We no longer have a 'root' predicate.  Items either go in the `Contract` or one of the named `Predicate`s.

In general this has simplified the code as there use to be quite bit of special casing around the root predicate.  It has also removed the `check_contract()` validation altogether as it was only checking for invalid items in the root predicate.

The type checking for paths was already fairly complex and now it's probably a little bit worse.  Paths can point to just about anything and so determining whether they're valid is tricky.  This code in `type_check.rs` could be further refactored for clarity.


otrho:
There were quite a few tests which were inadvertently abusing the root predicate by parsing expressions into it.  This isn't possible any more, so now those tests need to parse predicates containing expressions.

Ideally one day we stop parsing directly to the `Contract` and instead have an AST again.  All the parser tests could be against the AST rather than whatever ends up in the `Contract` or `Predicate`s.  And also much of the macro stuff would be simplified, so it's definitely a worthwhile refactor for one day.

mohammadfawaz (approved):

PR #803, opened 2024-07-30
Storage vectors and `__vec_len` intrinsic

mohammadfawaz:
Adds a storage vector type that can only be used in storage.
* New intrinsic `__vec_len` that returns the length of a vector and should only be used as a storage access, i.e., to initialize a `state` variable. I expect this intrinsic to be wrapped in a `macro` when we have a standard library.
* A storage vector stores its length in storage, at the base key.
* Keys in a storage vectors are assigned similarly to storage maps.
* Add OOB checks in the flattening phase.
* **Currently, only 1-dimensional vectors of basic types are supported.**

### DIFF
--- a/docs/book/src/appendix/intrinsics.md
+++ b/docs/book/src/appendix/intrinsics.md
@@ -67,6 +67,14 @@ variable or a "next state" expression but can have any type.
 ---
 
 ```pint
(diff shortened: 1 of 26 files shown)


otrho (approved):
Nice.

Issue #660, opened 2024-06-18
Type errors are a little excessive.


otrho:
This is mostly a low priority nice-to-have but error recovery during type checking will report multiple errors for the same line when one really is enough.

The most obvious redundancy is something like this:
```
var a: int = 1 + false;
```
produces something like... (this might be out of date, but it approximates the problem...)
```
Error: binary operator type error
   ╭─[foo.pnt:1:18]
   │
 1 │ var a: int = 1 + false;
   │                  ──┬──
   │                    ╰──── operator `+` argument has unexpected type `bool`
───╯
Error: variable initialization type error
   ╭─[foo.pnt:1:14]
   │
 1 │ var a: int = 1 + false;
   │        ─┬─   ────┬────
   │         ╰─────────────── expecting type `int`
   │                  │
   │                  ╰────── variable initializer has unknown type
───╯
Error: constraint expression type error
   ╭─[foo.pnt:1:1]
   │
 1 │ var a: int = 1 + false;
   │ ───────────┬──────────
   │            ╰──────────── constraint expression has unknown type
   │            │
   │            ╰──────────── expecting type `bool`
───╯
```
The last one is expecially redundant.  I'm not sure what the solution is since we don't want to suppress unrelated messages which still may be on the same line.  Perhaps a single type error specifically per line is enough though.

PR #811, opened 2024-07-31
Excessive error messages related to intrinsics

### LINKED ISSUE #804: Excessive error messages related to intrinsics
Here's an example:
```rust
storage {
    x: int,
}

predicate Test {
    state x = storage::x;
    var len = __state_len(x, 0);
}
```
I'm looking specifically at the first two errors below which are identical.
```console
❯ pintc test.pnt
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: unable to determine expression type
   ╭─[test.pnt:7:9]
   │
 7 │     var len = __state_len(x, 0);
   │         ─┬─
   │          ╰─── type of this expression is ambiguous
───╯
Error: constraint expression type error
   ╭─[test.pnt:7:5]
   │
 7 │     var len = __state_len(x, 0);
   │     ─────────────┬─────────────
   │                  ╰─────────────── constraint expression has unknown type
   │                  │
   │                  ╰─────────────── expecting type `bool`
───╯
Error: could not compile `test.pnt` due to 4 previous errors
```

To make things worse, the following:
```rust
storage {
    x: int,
}

predicate Test {
    state x = storage::x;
    var len = __state_len(x, 0);

    constraint len == 0;
    constraint len == 0;
    constraint len == 0;
    constraint len == 0;
    constraint len == 0;
    constraint len == 0;
    constraint len == 0;
}
```
results in even more repetition:
```console
❯ pintc test.pnt
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: unable to determine expression type
   ╭─[test.pnt:7:9]
   │
 7 │     var len = __state_len(x, 0);
   │         ─┬─
   │          ╰─── type of this expression is ambiguous
───╯
Error: constraint expression type error
   ╭─[test.pnt:7:5]
   │
 7 │     var len = __state_len(x, 0);
   │     ─────────────┬─────────────
   │                  ╰─────────────── constraint expression has unknown type
   │                  │
   │                  ╰─────────────── expecting type `bool`
───╯
Error: constraint expression type error
   ╭─[test.pnt:9:5]
   │
 9 │     constraint len == 0;
   │     ─────────┬─────────
   │              ╰─────────── constraint expression has unknown type
   │              │
   │              ╰─────────── expecting type `bool`
───╯
Error: constraint expression type error
    ╭─[test.pnt:10:5]
    │
 10 │     constraint len == 0;
    │     ─────────┬─────────
    │              ╰─────────── constraint expression has unknown type
    │              │
    │              ╰─────────── expecting type `bool`
────╯
Error: constraint expression type error
    ╭─[test.pnt:11:5]
    │
 11 │     constraint len == 0;
    │     ─────────┬─────────
    │              ╰─────────── constraint expression has unknown type
    │              │
    │              ╰─────────── expecting type `bool`
────╯
Error: constraint expression type error
    ╭─[test.pnt:12:5]
    │
 12 │     constraint len == 0;
    │     ─────────┬─────────
    │              ╰─────────── constraint expression has unknown type
    │              │
    │              ╰─────────── expecting type `bool`
────╯
Error: constraint expression type error
    ╭─[test.pnt:13:5]
    │
 13 │     constraint len == 0;
    │     ─────────┬─────────
    │              ╰─────────── constraint expression has unknown type
    │              │
    │              ╰─────────── expecting type `bool`
────╯
Error: constraint expression type error
    ╭─[test.pnt:14:5]
    │
 14 │     constraint len == 0;
    │     ─────────┬─────────
    │              ╰─────────── constraint expression has unknown type
    │              │
    │              ╰─────────── expecting type `bool`
────╯
Error: constraint expression type error
    ╭─[test.pnt:15:5]
    │
 15 │     constraint len == 0;
    │     ─────────┬─────────
    │              ╰─────────── constraint expression has unknown type
    │              │
    │              ╰─────────── expecting type `bool`
────╯
Error: could not compile `test.pnt` due to 18 previous errors
```

lethalgem:
Closes #804

This PR no longer checks an `expr_key` that failed type inference. Previously we were checking a failed `expr_key` multiple times and adding an error to the `handler` each time. 

The happy path where we successfully infer the `type` of an `expr_key` remains unchanged and each subsequent check of the `expr_key` will be skipped as before.

To accomplish skipping, I've reported the `type` of the `expr_key` as `Type::Error`. Previously this was only ever assigned when we have an error during `parsing`. I believe this is still an appropriate use of the `Type::Error`. Please let me know if that is not correct.

This has removed a lot of redundant errors in our testing. 

It has also had the side effects of:

Reporting failed `type_checks` as `type Error` instead of unknown
```
- // @436..473: constraint expression has unknown type
+ // @436..473: constraint expression has unexpected type `Error`
```

Reporting some _but not all_ types as `ambiguous`
```
- // unable to determine expression type
- // @60..61: type of this expression is ambiguous
```


mohammadfawaz:
Two questions:
1. Have you checked that all the error messages you removed from the tests are indeed unnecessary?
2. Does this PR also close https://github.com/essential-contributions/pint/issues/660?

mohammadfawaz:
Also, `unknown type` is better than `unexpected type `Error`. What did this change and what can we do about it?

lethalgem:
> Also, `unknown type` is better than `unexpected type `Error`. What did this change and what can we do about it?

Assigning a failed `type_check` to `Type::Unknown` caused the change in error messages.

This is because we check the type of `expr_key` only if it is `Type::Unknown`. So assigning the type to `unknown` when it fails would not solve the issue, we would just continue to try and `type_check` the `expr_key`.

```rust
if !next_key.get_ty(self).is_unknown() {
                queue.pop();
            } else {
                match self
                    .infer_expr_key_type(pred_key.map(|pred_key| &self.preds[pred_key]), next_key)
                {
.
.
.
                    // Some exprs may fail to be inferred. In that case we need to mark the
                    // expr_key as failing so that we don't check it again and report the
                    // error more than once.
                    Err(inference_error) => {
                        next_key.set_ty(Type::Error(inference_error.span().clone()), self);
                        return Err(inference_error);
                    }
                }
            }
```

Depending on how we want to use the error message for `Type::Error`, I can add to `generate_type_error_labels()` to treat `Type::Error` as `unknown` as well. Both statements are true and provide the same amount of information to the user.

https://github.com/essential-contributions/pint/blob/7dadd1fc5cc5f1afa0723b02ef45b8b741f225b3/pintc/src/error/compile_error.rs#L1259-L1270

mohammadfawaz:
> I can add to `generate_type_error_labels()` to treat `Type::Error` as `Unknown` as well. Both statements are true and provide the same amount of information to the user.

That's a good solution yes.

lethalgem:
> Two questions:
> 
> 1. Have you checked that all the error messages you removed from the tests are indeed unnecessary?
> 2. Does this PR also close [Type errors are a little excessive. #660](https://github.com/essential-contributions/pint/issues/660)?

1. The error messages are unnecessary for the user, though they are legitimate errors that are no longer reported. Ex. below, the removed lines are a real error, but they are redundant in that it another error on top of the same span. Basically extra errors are being report similar to issue #660.

We report the error "type of this expression is ambiguous" when the type is unknown for `var_inits` and `state_inits` later in `type_check_all_exprs`. We lose resolution with other errors as well because of the early assign to `Type::Error` instead of `Type::Unknown` when erroring on `type inference`. From what I saw in the changelogs, they are also redundant because the error that caused the failed `type inference` is already logged and reported for that span.

```
predicate Foo {
    state y = storage::y;
}
// parsed <<<
// predicate ::Foo {
//     state ::y = storage::y;
// }
// >>>
// typecheck_failure <<<
// cannot find storage variable `y`
// @30..40: no storage declaration found
- // unable to determine expression type
- // @20..40: type of this expression is ambiguous
// >>>
```

3. Unfortunately it does not. The example in the issue #660 does still result in the 2 redundant messages.

mohammadfawaz (approved):

otrho:
Yeah, the 'is ambiguous' error should really be a last resort error if there are no other errors for a particular failure.  But I think in practise that is pretty much never.  I was looking at it recently and thinking it should only be emitted if the `Handler` is empty, but that would mean having a separate `Handler` for every expression or something.  IDK.  Low priority.

PR #813, opened 2024-08-01
Disallow state variables from having types that contain storage types #805

### LINKED ISSUE #805: Disallow `state` variables from having types that contain storage types
The following should error out:
```rust
storage {
    t: { ( int => int ), int },
}

predicate test {
    state t = storage::t;
}
```
mainly because it's unclear how to refer to a storage map from a `state` variable. We already disallow state variables of type `Map` but we should also disallow state variables of types that _contain_ maps (or other storage type such as storage vector).

lethalgem:
Closes #805 

This PR reports an error if the `state_ty` or any nested `type` is a `map` or a `vector`

It's worth noting the error message states the `map` or `vector` type that is found. I _think_ this is helpful because the user can go down the rabbit hole of nested types and know which one is causing the issue.

The alternative would be to list the `state_ty` in the error and then the user would need to check what types are nested until they come across a `map` or `vector`.

Ex.
```pnt
storage {
    t: { ( int => int ), int },
}

predicate test {
    state t = storage::t;
}

Error: state variables cannot have storage types
   ╭─[test.pnt:7:5]
   │
 7 │     state t = storage::t;
   │     ──────────┬─────────  
   │               ╰─────────── found state variable of storage type ( int => int ) here
───╯
```


lethalgem:
Should this be allowed / should this be a separate issue?

```pnt
storage {
    x: ( int => int ),
    t: { x, int },
}

predicate test {
    state t = storage::t;
}
```

We do not report any compilation arrows for the above. `::x` is not considered a `new type decl,` but is considered to be a `custom` type. So when checking for a map it fails. Though I think this should be a syntactic error because `::x` should be accessed as `storage::x` which isn't allowed inside of `storage{}`

mohammadfawaz:
> Should this be allowed / should this be a separate issue?
> 
> ```
> storage {
>     x: ( int => int ),
>     t: { x, int },
> }
> 
> predicate test {
>     state t = storage::t;
> }
> ```
> 
> We do not report any compilation arrows for the above. `::x` is not considered a `new type decl,` but is considered to be a `custom` type. So when checking for a map it fails. Though I think this should be a syntactic error because `::x` should be accessed as `storage::x` which isn't allowed inside of `storage{}`

Yeah this should fail somewhere else because `x` is not a type. What you're trying to do shouldn't work because `x` is a storage variable while `{ x, int }` is a type.

lethalgem:
> > Should this be allowed / should this be a separate issue?
> > ```
> > storage {
> >     x: ( int => int ),
> >     t: { x, int },
> > }
> > 
> > predicate test {
> >     state t = storage::t;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We do not report any compilation arrows for the above. `::x` is not considered a `new type decl,` but is considered to be a `custom` type. So when checking for a map it fails. Though I think this should be a syntactic error because `::x` should be accessed as `storage::x` which isn't allowed inside of `storage{}`
> 
> Yeah this should fail somewhere else because `x` is not a type. What you're trying to do shouldn't work because `x` is a storage variable while `{ x, int }` is a type.

Made it into a new issue

pintc/tests/storage/invalid_state_var_storage_type.pnt
```diff
@@ -0,0 +1,31 @@
...
+//
+// predicate ::test {
+//     state ::d = storage::b;
+//     state ::e = storage::c;
+// }
+// >>>
+
+// typecheck_failure <<<
+// state variables cannot have storage types
+// @122..142: found state variable of storage type ( int => int ) here
```
mohammadfawaz, line 28: so the error is a little misleading.. The type of the stat variable is not `( int => int )`. It's the whole tuple. However, the tuple itself contains the storage type. So we should reword this. I know you reused the old error variant `StateVarHasStorageType` but I don't think it fits anymore.

lethalgem: Just updated. Instead of changing the error, I changed it so that we print the `parent_ty` if any `children_ty` are found to be a `map` or `vector`.

Ex. 
```
Error: state variables cannot have storage types
   ╭─[test.pnt:9:5]
   │
 9 │     state d = storage::b;
   │     ──────────┬─────────  
   │               ╰─────────── found state variable of storage type {( int => int ), int} here
───╯
Error: state variables cannot have storage types
    ╭─[test.pnt:10:5]
    │
 10 │     state e = storage::c;
    │     ──────────┬─────────  
    │               ╰─────────── found state variable of storage type {::a ({( bool => bool ), int}), int} here
────╯
```

Let me know if that's more appropriate or if you have something else in mind.

lethalgem: In terms of rewording instead of printing the `parent_ty`, what do you think of:
```
Error: state variables cannot have storage types
   ╭─[test.pnt:9:5]
   │
 9 │     state d = storage::b;
   │     ──────────┬─────────  
   │               ╰─────────── state variable cannot have nested type of ( int => int )
───╯
```

I can then differentiate between the errors based on if its the first pass or subsequent ones.

mohammadfawaz: No what you have is fine. However, maybe adding a note would be helpful? Something like:
```
Note:  type of state variable is `{( int => int ), int}` which depends on the storage type `( int => int )`.
```

lethalgem:
```
Error: state variables cannot have storage types
   ╭─[test.pnt:9:5]
   │
 9 │     state d = storage::b;
   │     ──────────┬─────────  
   │               ╰─────────── found state variable of storage type {( int => int ), int} here
   │ 
   │ Note: type of state variable is `{( int => int ), int}` which depends on the storage type `( int => int )`.
───╯
Error: state variables cannot have storage types
    ╭─[test.pnt:10:5]
    │
 10 │     state e = storage::c;
    │     ──────────┬─────────  
    │               ╰─────────── found state variable of storage type {::a ({( bool => bool ), int}), int} here
    │ 
    │ Note: type of state variable is `{::a ({( bool => bool ), int}), int}` which depends on the storage type `( bool => bool )`.
────╯
```
seems a bit redundant to me. What do you think of
```
Error: state variables cannot have storage types
   ╭─[test.pnt:9:5]
   │
 9 │     state d = storage::b;
   │     ──────────┬─────────  
   │               ╰─────────── found state variable of storage type {( int => int ), int} here
   │ 
   │ Note: type of state variable is `{( int => int ), int}` which depends on the storage type `( int => int )`.
───╯
Error: state variables cannot have storage types
    ╭─[test.pnt:10:5]
    │
 10 │     state e = storage::c;
    │     ──────────┬─────────  
    │               ╰─────────── found state variable of storage type {::a ({( bool => bool ), int}), int} here
    │ 
    │ Note: type of state variable depends on the storage type `( bool => bool )`.
────╯
```

mohammadfawaz: Yeah that's reasonable

mohammadfawaz (approved):

Issue #812, opened 2024-07-31
More work around excessive error messages


mohammadfawaz:
E.g.
```rust
storage {
    x: int,
}

predicate Test {
    state x = storage::x;
    var len = __state_len(x, 0);

    constraint len == 0;
}
```
we get 
```
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: binary operator type error
   ╭─[test.pnt:9:23]
   │
 9 │     constraint len == 0;
   │                ─┬─    ┬
   │                 ╰──────── expecting type `Error`
   │                       │
   │                       ╰── operator `==` argument has unexpected type `int`
───╯
Error: constraint expression type error
   ╭─[test.pnt:7:5]
   │
 7 │     var len = __state_len(x, 0);
   │     ─────────────┬─────────────
   │                  ╰─────────────── constraint expression has unknown type
   │                  │
   │                  ╰─────────────── expecting type `bool`
───╯
Error: constraint expression type error
   ╭─[test.pnt:9:5]
   │
 9 │     constraint len == 0;
   │     ─────────┬─────────
   │              ╰─────────── constraint expression has unknown type
   │              │
   │              ╰─────────── expecting type `bool`
───╯
Error: could not compile `test.pnt` due to 4 previous errors
```
There are few issues:
1. `expecting type 'Error'` in the second error is weird.
2. Ideally, if an error refers to "Unknown" or "Error" such as the other one, we shouldn't emit it. It's not very useful and is not actionable.

otrho:
(2) is a desirable fix.  I just came across a `illegal cast from a 'Error'`.  Kinda crap.

lethalgem:
Note to self:
```
Error: constraint expression type error
   ╭─[test.pnt:7:5]
   │
 7 │     var len = __state_len(x, 0);
   │     ─────────────┬─────────────
   │                  ╰─────────────── constraint expression has unknown type
   │                  │
   │                  ╰─────────────── expecting type `bool`
───╯

is because we didn't infer the type for

Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯

[...]

PR #820, opened 2024-08-06
Assign `intrinsic` types even if malformed

lethalgem:
Just the start to #812. This PR significantly cuts down on chained errors due to errors inferring `intrinsics`.

Before, if we saw a proper `intrinsic` name (ex. __sha256 with the wrong # of arguments) we would `handler.emit_err()` and bail on inferring the type. This caused extra errors down the line because further `type_checking` operated on a `Type::Error | Type::Unknown` instead of `Type::Int` or whatever the type of the malformed `intrinsic call` was.

By assigning the proper `type` to the `intrinsic call` the rest of the `type_checking` goes smoothly and prevents a lot of future errors that are not helpful.

My thought, after having discussed with @mohammadfawaz, is that we shouldn't bail early near as often as we currently do when inferring a type. Typically we know what the type _should be_. If we assign that type, emit the error, and move on then we would prevent a lot of errors from cropping up in the subsequent `type_check` stages. This also helps the user know the root cause of the issue instead of sifting through 4+ errors to figure out which one is the actionable one.

### DIFF
--- a/pintc/src/predicate/analyse/intrinsics.rs
+++ b/pintc/src/predicate/analyse/intrinsics.rs
@@ -31,9 +31,9 @@ impl Contract {
                 // Access ops
                 "__mut_keys_len" => infer_intrinsic_mut_keys_len(args, span),
                 "__mut_keys_contains" => infer_intrinsic_mut_keys_contains(self, name, args, span),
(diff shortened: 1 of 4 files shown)


mohammadfawaz (approved):
Looks good to me!

PR #822, opened 2024-08-08
Mutable keys and refactoring storage

### LINKED ISSUE #797: Add a `mut` modifier to `storage::..` expressions
As a way to describe mutable keys...
.. TODO

mohammadfawaz:
1. Completely remove `__mut_keys_len` and `__mut_keys_contains` since the VM doesn't have the corresponding opcodes anymore.
2. Remove `__storage_get` and `__storage_get_extern` from type checking. They're only allowed to be inserted by the compiler and handled in asm gen.
3. Add `__eq_set` and `__mut_keys` but only allow them to be added by the compiler, similarly to the storage intrinsics above
4. `mut` keyword for storage accesses. External storage accesses don't have that since external storage keys should be constrained in the contract that owns them.
5. Moved key computation for storage variables into a lowering pass. The same pass also insert the constraint
```rust
constraint __eq_set(__mut_keys(), { /* list of keys */ })
```
Closes #797


otrho:
fileschanged

Yikes!

mohammadfawaz:
> fileschanged
> 
> Yikes!

Well, every test gets a new constraint on mutable keys, even when there are none.

otrho (approved):

PR #824, opened 2024-08-09
Cut down unnecessary errors from failed inferences

### LINKED ISSUE #660: Type errors are a little excessive.

lethalgem:
Continuation of the #812 refactor. Specifically builds on the pattern of #820. It also closes #660.

In summary, we now do our best to `infer` an appropriate `type` for an `expr` even if the `expr` has an error. This is has dramatically cut down the amount of errors being reported.

Before we were exiting the inference process the moment that an error was discovered. This would cause a cascade effect where subsequent errors were generated because we were now analyzing a `Type::Error` or `Type::Unknown`. A user would sometimes have to sort through 4+ error messages to understand which was the one they could action on. By assigning an appropriate type, only the actionable root cause error is reported.

It has also had the unintended side effect of surfacing some actionable error messages that were previously not generated because we exited the process early.

-----

As a small example, take issue #660
```
predicate Test {
    var a: int = 1 + false;
}
```
Previously produced the 3 errors below. Only the first error is actionable to the user and resolving it would resolve all the other errors.
```
Error: binary operator type error
   ╭─[foo.pnt:1:18]
   │
 1 │ var a: int = 1 + false;
   │                  ──┬──
   │                    ╰──── operator `+` argument has unexpected type `bool`
───╯
Error: variable initialization type error
   ╭─[foo.pnt:1:14]
   │
 1 │ var a: int = 1 + false;
   │        ─┬─   ────┬────
   │         ╰─────────────── expecting type `int`
   │                  │
   │                  ╰────── variable initializer has unknown type
───╯
Error: constraint expression type error
   ╭─[foo.pnt:1:1]
   │
 1 │ var a: int = 1 + false;
   │ ───────────┬──────────
   │            ╰──────────── constraint expression has unknown type
   │            │
   │            ╰──────────── expecting type `bool`
───╯
```
Now it produces...
```
Error: binary operator type error
     ╭─[test.pnt:189:22]
     │
 189 │     var a: int = 1 + false;
     │                      ──┬──  
     │                        ╰──── operator `+` argument has unexpected type `bool`
─────╯
```

-----

Another example from issue #812
```
storage {
    x: int,
}

predicate Test {
    state x = storage::x;
    var len = __state_len(x, 0);

    constraint len == 0;
}
```
Previously produced 4 errors. 3 of which were caused by the first error.
```
Error: this intrinsic takes 1 argument but 2 arguments were supplied
   ╭─[test.pnt:7:15]
   │
 7 │     var len = __state_len(x, 0);
   │               ────────┬────────
   │                       ╰────────── unexpected number of arguments here
───╯
Error: binary operator type error
   ╭─[test.pnt:9:23]
   │
 9 │     constraint len == 0;
   │                ─┬─    ┬
   │                 ╰──────── expecting type `Error`
   │                       │
   │                       ╰── operator `==` argument has unexpected type `int`
───╯
Error: constraint expression type error
   ╭─[test.pnt:7:5]
   │
 7 │     var len = __state_len(x, 0);
   │     ─────────────┬─────────────
   │                  ╰─────────────── constraint expression has unknown type
   │                  │
   │                  ╰─────────────── expecting type `bool`
───╯
Error: constraint expression type error
   ╭─[test.pnt:9:5]
   │
 9 │     constraint len == 0;
   │     ─────────┬─────────
   │              ╰─────────── constraint expression has unknown type
   │              │
   │              ╰─────────── expecting type `bool`
───╯
Error: could not compile `test.pnt` due to 4 previous errors
```
Now it produces
```
Error: this intrinsic takes 1 argument but 2 arguments were supplied
     ╭─[test.pnt:198:15]
     │
 198 │     var len = __state_len(x, 0);
     │               ────────┬────────  
     │                       ╰────────── unexpected number of arguments here
─────╯
```


lethalgem:
This may also even close issue #812

That is up to ya'll to decide as the reviewers.

pintc/tests/arrays/bad_const_index.pnt
```diff
@@ -17,7 +17,7 @@ predicate test {
 // attempt to index an array with a mismatched value
 // @71..75: array access must be with an int value
 // found access using type `real`
-// constraint expression type error
-// @58..81: constraint expression has unknown type
-// @58..81: expecting type `bool`
+// binary operator type error
+// @80..81: operator `==` argument has unexpected type `int`
+// @69..76: expecting type `real`
```
mohammadfawaz, line 22: This shouldn't be an error right? `a[1.2]` is an `int` technically even though the index is wrong... Saying that we expect type `real` seems incorrect, no?

pintc/src/predicate/analyse/type_check.rs
```diff
@@ -881,56 +867,65 @@ impl Contract {
         Ok(())
     }
 
+    // TODO: in refactor, make this return error emitted and not error
```
mohammadfawaz, line 870: This `TODO` is no longer needed?

pintc/src/predicate/analyse/type_check.rs
```diff
@@ -1987,89 +2044,97 @@ impl Contract {
...
     fn infer_generator_expr(
         &self,
+        handler: &Handler,
         kind: &GeneratorKind,
         ranges: &[(Ident, ExprKey)],
         conditions: &[ExprKey],
         body_expr_key: ExprKey,
         span: &Span,
-    ) -> Result<Inference, Error> {
+    ) -> Inference {
```
mohammadfawaz, line 2130: So I would really prefer if all these methods return `Result<Inference, ErrorEmitted>` instead of just `Inference`.

pintc/src/predicate/analyse/type_check.rs
```diff
@@ -570,12 +566,10 @@ impl Contract {
                 .collect::<Vec<_>>()
                 .iter()
             {
-                if let Err(err) = self.type_check_single_expr(handler, Some(pred_key), *range_expr)
-                {
-                    handler.emit_err(err);
-                } else if !(range_expr.get_ty(self).is_int()
-                    || range_expr.get_ty(self).is_enum(&self.enums)
-                    || checked_range_exprs.contains(range_expr))
+                if self.type_check_single_expr(handler, Some(pred_key), *range_expr) == Ok(())
```
mohammadfawaz, line 569: instead of `== Ok(())`, `is_ok()` should work.

pintc/src/predicate/analyse/type_check.rs
```diff
@@ -944,139 +939,153 @@ impl Contract {
...
+            Expr::Range { lb, ub, span } => Ok(self.infer_range_expr(handler, *lb, *ub, span)),
 
             Expr::Generator {
                 kind,
                 gen_ranges,
                 conditions,
                 body,
                 span,
-            } => self.infer_generator_expr(kind, gen_ranges, conditions, *body, span),
+            } => Ok(self.infer_generator_expr(handler, kind, gen_ranges, conditions, *body, span)),
```
mohammadfawaz, line 968: What's a little weird here is that we may be returning `Ok(..)` on something that may fail. Ideally, you want `infer_generator_expr` to return a `Result<_, ErrorEmitted>` and just bubble that up. It's a big change to what you have but I think it's more robust. That way, we can ask the question: "did this method fail?".

Also, for that to work well, you need a scoped handler:
```
handler.scope(|handler| infer_generator_expr(handler, ...))
```
and then, in `infer_generator_expr`, you would add at the end:
```
        if handler.has_errors() {
            Err(handler.cancel());
        } else {
            Ok(())
        }
```
something like that...

otrho: > That way, we can ask the question: "did this method fail?".

Or we can ignore it if we want.  Flexibility both ways.  Sometimes we can just ask the handler `.has_errors()`? but that isn't necessarily specific enough for all cases.

```
  if handler.has_errors() {
    Err(handler.cancel());
  } else {
    Ok(())
  }
```
We use this often enough that it should probably be a method on `Handler` IMO.

lethalgem: After talking with @mohammadfawaz we've determined it isn't possible to return `Result<Inference, ErrorEmitted>` with the current type inference structure. This PR cuts down on unnecessary errors by doing both, returning an `Inference` and emitting an error through `handler.emit_err()`.

We talked about changing the structure to return a `Result<(), ErrorEmitted>` and having `set_ty()` at the end of each inference branch. This would let us both set the type to be the `Inference` and return `ErrorEmitted` if there was a malformed `expr`. However this fails when we have `Inference::Dependant` as we need to bubble the `Inference` up to `type_check()` so that we can check the type of the `dependant expr` first.

In the end we decided to leave the structure as it is and if we feel we need to react to an `err_emit()` in the future, we can address restructuring then.

pintc/src/predicate/analyse/type_check.rs
lethalgem: Good catch! I was inferring the wrong type in array accesses. This also helped remove even more cascading errors in other tests.

pintc/tests/arrays/index_enum_with_int.pnt
```diff
@@ -19,7 +19,7 @@ predicate test {
 // attempt to index an array with a mismatched value
 // @98..99: array access must be with a `::Num` variant
 // found access using type `int`
-// constraint expression type error
-// @83..105: constraint expression has unknown type
-// @83..105: expecting type `bool`
+// binary operator type error
+// @104..105: operator `==` argument has unexpected type `int`
+// @94..100: expecting type `::Num`
```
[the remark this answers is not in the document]
lethalgem: Another good catch. It's been corrected :)

otrho (approved):
Yep, I'm happy with this approach.  The `if let Err(e) = ... { handler.emit_err(e); }` stuff was crap and the check for `.is_ok()` before doing further checks is a good way to restrict the error output.

`Result` and `?` just aren't designed for error recovery and `ErrorEmitted` is half a solution but the best we have for now.

PR #826, opened 2024-08-12
Insert constraint on pathway addresses and refactor storage

### LINKED ISSUE #646: Implement short-circuiting using `JumpForwardIf`
Example:
```rust
constraint A || B;
```
is equivalent to
```rust
if A {
    true 
} else {
    B
}
```
while we do convert `if` back to an OR, what we may want to do in codegen is:
* Evaluate `A`
* Jump to evaluating `B` _if_ `A` is false. This uses `JumpForwardIf`.

mohammadfawaz:
1. Refactor transient key and pathway computation out of asm gen into a separate lowering pass
2. Insert constraints on the addresses used in contract and predicate instances. These should match what `__predicate_at` returns.

Closes #646


pintc/src/asm_gen.rs
```diff
@@ -98,14 +96,44 @@ impl AsmBuilder {
...
             | Expr::Generator { .. }
             | Expr::StorageAccess { .. }
             | Expr::ExternalStorageAccess { .. } => Ok(Location::Value),
+            Expr::IntrinsicCall { name, args, .. } => {
+                if name.name == "__transient" {
+                    // This particular intrinsic returns a `Location::Transient`. All other
+                    // intrinsics are just values.
+
+                    // Expecting 3 arguments:
+                    assert_eq!(args.len(), 3);
```
otrho, line 112: Whenever I see an `assert` in here, particularly in a function which returns a `Result` I'm thinking it should be an `InternalError`.

It's debatable whether the `InternalError` is actually better though.  I feel like a 'crash' is bad for the user to see, but then an internal error is bad too.  Perhaps if we keep `assert`s they should at least have a message.

mohammadfawaz: Yeah we need to be consistent I agree, and we're not at the moment. Especially in asm gen. Internal compiler errors can be better if we make them more descriptive (e.g. where exactly the error happened etc.). Maybe we need a macro to wrap the 5-6 lines in takes to emit an internal error.

otrho: It also currently takes a `&str` description which means it can't be `format!(..)`ed.  Until now I haven't minded since as soon as I see an internal error I search the source for that message string and then start `dbg!()`ing or whatever.

otrho (approved):

PR #833, opened 2024-08-19
Do not compile dead state decls

### LINKED ISSUE #815: Do not compile dead state declarations
For example:
```rust
storage {
    x: int,
}

predicate Test {
    state x = storage::x;
}
```
`state x` is not used anywhere in the predicate so the compiler can just skip it completely when generating assembly. That being said, if the declaration had `mut` in it: `state x = mut storage::x`, then mutability should still be taken into account (i.e. should still generate a `__mut_keys_contains` constraint)

lethalgem:
Closes #815


pintc/src/predicate/transform/clean.rs
```diff
@@ -23,12 +25,21 @@ pub(crate) fn clean_dead_state_decls(contract: &mut Contract) {
...
+                // if dead_state.get(pred)
+                state_expr = Some(dead_state.get(pred).expr);
+
                 pred.states.remove(dead_state);
             }
         }
 
+        if let Some(expr) = state_expr {
+            println!("expr: {}", contract.with_ctrct(expr));
+        } // TODO: Find a way to deal with the states being mut. For now we are not ignoring them and will remove them if they are unused
```
mohammadfawaz, line 41: That's fine.. no need to worry about mutability. That's all handled in the pass `lower_storage_accesses`. By the time your pass runs, all of that stuff in consumed.

pintc/tests/interfaces/simple.pnt
```diff
@@ -52,6 +52,10 @@ predicate Simple {
     state x = FooInstance::storage::x;
     state y = FooInstance::storage::y;
     state z = BarInstance::storage::z;
+    constraint x == x;
+    constraint y == y;
+    constraint z == z;
+
```
mohammadfawaz, line 58: So I hate to make you do more work, but in the **near** future, we will almost definitely add a new pass that will optimize away trivial constraints like this. Therefore, it's probably a better idea to make these truly _not dead_. For example:
```rust
constraint x == 0;
constraint y == 0;
constraint z == 0;
```
and so on... 
for complex types like tuples and arrays, you can always do something like:
```
constraint t.0.1 == 0;
```
assuming `t.0.1` is an integer. That is you don't really need to constrain the whole complex variable.

pintc/src/predicate/transform/dead_code_elimination.rs
```diff
@@ -0,0 +1,39 @@
...
+
+/// In a given contract, remove any code that is not reachable or used.
+pub(crate) fn dead_code_elimination(contract: &mut Contract) {
+    dead_state_elimination(contract);
+}
+
+/// Remove all unused States in their respective predicates.
+pub(crate) fn dead_state_elimination(contract: &mut Contract) {
+    for pred_key in contract.preds.keys().collect::<Vec<_>>() {
+        let mut live_paths: Vec<String> = Vec::new();
```
mohammadfawaz, line 14: Since you don't care about the order here, it's more efficient to change this into a `HashSet`. You also know that paths _have_ to be unique.

pintc/src/predicate/transform/dead_code_elimination.rs
```diff
@@ -0,0 +1,39 @@
...
+                live_paths.push(name.to_string());
+            }
+        }
+
+        let pred_states = contract
+            .preds
+            .get(pred_key)
+            .expect("pred guaranteed to exist")
+            .states();
+        let mut dead_state_decls: Vec<StateKey> = Vec::new();
```
mohammadfawaz, line 26: `dead_state_decls` should probably be a `HashSet` as well.

pintc/src/predicate/transform/dead_code_elimination.rs
```diff
@@ -0,0 +1,39 @@
...
+        for state in pred_states {
```
mohammadfawaz, line 27: So this is a stylistic issue, but I prefer something like:
```
for (state_key, state) in pred_states {
    if !live_paths.contains(&state.name
            dead_state_decls.push(state_key);
    }
}
```
This is just much clearer than having to think about what `state.0` and `state.1` are.

pintc/src/predicate/transform/dead_code_elimination.rs
```diff
@@ -0,0 +1,39 @@
...
+
+/// Remove all unused States in their respective predicates.
+pub(crate) fn dead_state_elimination(contract: &mut Contract) {
+    for pred_key in contract.preds.keys().collect::<Vec<_>>() {
+        let mut live_paths: Vec<String> = Vec::new();
+        for expr in contract.exprs(pred_key) {
+            if let Expr::Path(name, _) = expr.get(contract) {
+                live_paths.push(name.to_string());
+            }
+        }
```
mohammadfawaz, line 19: If you want to practice some functional-style Rust, you can try to rewrite this using a `filter_map` and then `collect` into a `HashSet`.

mohammadfawaz (approved):
Much better, :shipit:

PR #835, opened 2024-08-20
General refactor of intrinsics and asm_gen

### LINKED ISSUE #823: Refactor intrinsics in the middleend to avoid string compares everywhere
We should really have an enum that represent all available intrinsics and `Expr::IntrinsicCall` should use that instead of the name of the intrinsic. This is more maintainable and make `match` statements on intrinisic kind more sensible.

mohammadfawaz:
Closes #823 

This also refactors asm gen in the process. I started refactoring intrinsics in asm gen and then got carried away. There should be nothing functional about this change.

The most important change is making the way we handle intrinsics data driven. We rely on methods in `intrinsics.rs` that tell us what the type of each intrinsic is and what argument(s) it expects.

The other big change is that we now handle all expressions in `compile_expr()` in asm gen. Previously, we had many different small functions that handled state v.s. transient v.s. regular expr. Anyways, this new design is much cleaner I believe. Most of the bulky code was also extracted to separate methods.

I had to add a new type called `Any` that indicates that a type can be anything. We don't have that in the language but it's useful for indicating that some intrinsics can take arguments of any type (such as `__sha256`). This may not be the best idea though so let me know what you think @otrho. The alternative is to have another `enum` that track intrinsic arguments types rules.


otrho (approved):
Yep, this is great, thanks!  🙂 

At first I wasn't sure about the `Type::Any` idea, especially in a statically typed language.  But using it only for intrinsics is better than the alternative.

I'm still not sure about equality though.  Should `Type::Any == Type::Error` be true?  Should even `Type::Error == Type::Error` be true?  Same with `Type::Unknown == Type::Unknown`, etc.

Ideally I think we'd have two `Type` enums, one for pre-type-checking where we can have unknowns and anys but after type checking we wouldn't have them nor errors... everything concrete and fixed.  🤔  One day.

PR #851, opened 2024-08-24
Remove dead `constraints`

### LINKED ISSUE #834: Remove dead constraints
For example, remove constraints like this:
```
constraint true;
constraint !false;
constraint 1 == 1;
constraint x == x;
```
and so on...

We can start with some simple stuff and evolve the pass to support more complex structures in the future.

### LINKED ISSUE #853: Optimize `constraints` by removing all if one evaluates to false
If any `constraints` evaluate to `false` just remove every constraint there is and add a `constraint false`. You know the program is going to fail every single time and every other `constraint` is useless. Effectively this is just merging all the `constraints` into the simplest form.

1. evaluate the constraints
2. the moment you find a false
3. exit and clear out the constraints vector
4. add a new constraint equal to false (look at how to add a expr of false, there are examples)
5. give it an empty span

lethalgem:
Closes #834 
Closes #853 

For now it only eliminates very trivial `constraints` that evaluate to an `immediate`. ex. `constraint true`.

This PR also introduces the new `optimisation` phase of the compiler. The phase is optional.

To skip optimisation when compiling use the flag `--skip-optimise`. To print the optimisation when compiling use the flag `--print-optimised`

To e2e test for the optimised output, just add `optimised >>>` at the end of the file.


pintc/tests/interfaces/simple.pnt
```diff
@@ -52,10 +52,6 @@ predicate Simple {
     state x = FooInstance::storage::x;
     state y = FooInstance::storage::y;
     state z = BarInstance::storage::z;
-    constraint x == 0;
```
mohammadfawaz, line 55: I don't think these are meant to be removed 🤔

pintc/src/main.rs
```diff
@@ -47,7 +47,18 @@ fn main() -> anyhow::Result<()> {
         }
     };
 
-    match handler.scope(|handler| compile_contract(handler, &flattened)) {
+    // Optimise
+    let contract = if args.skip_optimise {
+        flattened
+    } else {
+        let optimised = flattened.optimise();
```
mohammadfawaz, line 54: If you're going to separate `optimize` from `compile()` above, then you also need to call `optimise()` in `pink-pkg`.. search for where `compile(..)` is used.

pintc/tests/tests.rs
```diff
@@ -53,8 +53,11 @@ fn run_tests(sub_dir: &str) -> anyhow::Result<()> {
                 // Type check the parsed intent.
                 type_check(pred, &test_data, &mut failed_tests, &path))
             .and_then(|pred|
-                // Flatten the parsed intent check the result.
-                flatten_and_check(pred, &test_data, &mut failed_tests, &path));
+                // Flatten the parsed intent and check the result.
+                flatten_and_check(pred, &test_data, &mut failed_tests, &path))
+            .and_then(|pred|
+                // Optimise the flattened intent
+                optimise(pred, &test_data));
```
mohammadfawaz, line 60: I think we discussed adding a separate step and a separate section for optimized code in the tests. This doesn't seem to be doing that.

pintc/src/main.rs
mohammadfawaz: In fact, I think `optimize` should be part of `compile()`.. just make it easier. In tests, you can separate the steps which is what we do (`type_check`, then `flatten`, then `optimize`)

lethalgem: Do we not want the user to be able to skip the optimise step when compiling? Should it only be optional for tests?

pintc/tests/tests.rs
lethalgem: It is separated in `optimise()` based on whether or not the test has `optimise >>>`. If it's not present we skip the step.

```rust
fn optimise(pred: Contract, test_data: &TestData) -> Option<Contract> {
    let optimised = pred.optimise();
    if let Some(expected_optimised_str) = &test_data.optimised {
        similar_asserts::assert_eq!(expected_optimised_str.trim(), format!("{optimised}").trim());
        Some(optimised)
    } else {
        None
    }
}
```

pintc/tests/interfaces/simple.pnt
lethalgem: Are you saying for all of the constraints that I put in for PR #833, or just this particular instance in `simple.pnt`?

My understanding is that we don't want these because the constraints were added to avoid optimising out the dead states. I added these in commit 6780d484884a2e7e61d446aa4c3c13da07ce0cbe and set them to a nontrivial value in commit 84e9ca9e46e0300851a013cbfd415b2cf72ee92b.

pintc/src/main.rs
otrho: Good question.  I'd say optimisation shouldn't be optional for regular compiling, although it would be useful for development -- for us.  Perhaps a hidden flag to disable optimisations?  Are hidden options possible with `clap`?

lethalgem: `hide` is in the [documentation](https://docs.rs/clap/latest/clap/struct.Arg.html#method.hide) for `arguments`. I haven't found anything on hiding `options`.

I spent some time working on hiding `skip-optimise` but I couldn't get it to work. I can hide any `help` text for the option, but not the option itself.

ex. 
```rust
#[arg(long = "skip-optimise", help="skip optimisation phase", hide=true)]
    pub skip_optimise: bool,
```
just hides "skip optimisation phase".

```
Usage: pintc [OPTIONS] <FILEPATH>

Arguments:
  <FILEPATH>  

Options:
  -o, --output <OUTPUT>  
      --print-parsed     
      --print-flat       
      --print-asm        
      --print-optimised  
  -h, --help             Print help
```

otrho: Isn't this what we want?  I don't see `--skip-optimise' in the output there.

BTW, I think for docs we're meant to use the 'ize' suffix.  I've accidentally let a couple of 'ise's through by habit.

lethalgem: Well that's embarrassing. Yes, it is working and I kept reading `--print-optimised` as `--skip-optimised`. Will add it now.

> BTW, I think for docs we're meant to use the 'ize' suffix. I've accidentally let a couple of 'ise's through by habit.

haha! I figured we'd keep in the `ise` in honor of you being the first to say we should make it a new phase. I'll also update it to `ize`

mohammadfawaz (approved):
:shipit:

PR #857, opened 2024-09-03
Fix a few issues with new types and clean up various restrictions in the compiler

### LINKED ISSUE #814: Declaring a `custom type` in the `storage block` should error
> Should this be allowed / should this be a separate issue?
> 
> ```
> storage {
>     x: ( int => int ),
>     t: { x, int },
> }
> 
> predicate test {
>     state t = storage::t;
> }
> ```
> 
> We do not report any compilation arrows for the above. `::x` is not considered a `new type decl,` but is considered to be a `custom` type. So when checking for a map it fails. Though I think this should be a syntactic error because `::x` should be accessed as `storage::x` which isn't allowed inside of `storage{}`

Yeah this should fail somewhere else because `x` is not a type. What you're trying to do shouldn't work because `x` is a storage variable while `{ x, int }` is a type.

_Originally posted by @mohammadfawaz in https://github.com/essential-contributions/pint/issues/813#issuecomment-2263617754_

### LINKED ISSUE #838: Cleanup supported types in storage and transient data (and their error messages)
We're currently supporting specific types in storage and transient data which is fine. However, the error messages we emit if the types are not supported are not great. They are mostly internal errors. We need to clean all of that stuff up. Ideally, everything is supported including strings. Otherwise, clear error messages should be emitted that tell us why the type isn't supported, and hopefully it's because it's not possible (and not because we're lazy 😆 )

### LINKED ISSUE #854: Storage variable `Custom` types are not properly checked
E.g.
```rust
storage {
    x: Foo,
}
```
Errors:
```console
❯ pintc all_types.pnt
Error: compiler internal error: only ints, bools, and maps are currently allowed in a storage block
   ╭─[all_types.pnt:7:5]
   │
 7 │     x: Foo,
   │     ────┬───
   │         ╰───── only ints, bools, and maps are currently allowed in a storage block
───╯
```
We should instead be emitting an error that indicates that the type `Foo` does not exist. If we were to write this instead:
```rust
predicate Foo {
    var x: Foo;
}
```
we would get
```console
Error: undefined type
   ╭─[all_types.pnt:7:12]
   │
 7 │     var x: Foo;
   │            ─┬─
   │             ╰─── type is undefined
───╯
```
which is reasonable.

mohammadfawaz:
Closes #854 
Closes #838
Closes #814 

- Remove string literals and string type from the parser. They're still there in the lexer and the rest of the compiler. When we revisit strings, we can decide what to keep and what to change. The only place we still use strings is in the intrinsic `__address_of` so I'm still allowing that as a special case.
- Guard `real` behind an experimental feature called `"experimental-types"`. Moved some tests around and enabled them only when the feature is on. Otherwise, we emit an error for both type `real` and real literals. Again, when we decide to support them properly, we will figure out what to do with them. I wanted to hide strings behind the same flag but things got pretty hairy in the parser so I gave up.. but that's probably not important anyways.
- Handle new types and undefined types check in a few missing places like consts, interfaces, etc (+ added tests) 
- Remove restrictions on pub vars except for the fact that they can't be storage only types. Similarly to regular decision variables.
- Cleanup the restrictions on storage types. We emit clear error messages now until we support everything.


pintc/src/predicate/transform/lower.rs
```diff
@@ -321,31 +322,71 @@ pub(crate) fn lower_aliases(contract: &mut Contract) {
         }
     }
 
+    // Replace `Type::Alias` in every var and state decl type in the contract
+    contract
+        .preds
+        .keys()
+        .collect::<Vec<_>>()
+        .iter()
```
otrho, line 330: Instead of collecting the keys into a new vector and iterator I think it would work to just use `.copied()` or at least `.cloned()`.

otrho (approved):

pintc/src/predicate/transform/lower.rs
mohammadfawaz: I could not make it work :(

PR #865, opened 2024-09-04
Emit `warnings`

### LINKED ISSUE #855: Add warnings to the `handler`
It's time we emit warnings while we're in the optimize phase

lethalgem:
Closes #855


pintc/src/predicate/optimize.rs
```diff
@@ -1,10 +1,12 @@
...
+
 impl super::Contract {
-    pub fn optimize(mut self) -> Self {
+    pub fn optimize(mut self, handler: &Handler) -> Self {
         // Remove dead code
-        dead_code_elimination(&mut self);
+        let _ = dead_code_elimination(&mut self, handler);
 
         self
     }
```
lethalgem, line 12: @mohammadfawaz I currently don't have `optimize()` return `WarningEmitted`. I didn't see any use for the result right now. We simply optimize, emit warnings if we have any, and move on. There isn't any conditional compilation from a warning being emitted.

Open to your thoughts on this ofc.

pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -1,22 +1,32 @@
...
+) -> Result<(), WarningEmitted> {
+    dead_state_elimination(contract, handler);
+    dead_constraint_elimination(contract, handler);
+
+    if handler.has_warnings() {
+        Err(handler.continue_with_warnings())
+    } else {
+        Ok(())
+    }
 }
```
lethalgem, line 26: I set it up to return `WarningEmitted` if we need it in the future

pintc/tests/optimizations/dead_state_elimination.pnt
```diff
@@ -136,4 +136,21 @@ predicate Test {
...
+// state is unused
+// @675..698: state is unused
+// state is unused
+// @704..729: state is unused
+// state is unused
+// @735..761: state is unused
+// state is unused
+// @767..793: state is unused
+// state is unused
+// @848..872: state is unused
```
lethalgem, line 155: @mohammadfawaz You'll want to look at this. I have `DeadStates` being emitted as a warning. I'm not familiar enough with how `mut` works to know if we should be letting the user know that the `state` is unused or not. Currently it is being reported as a warning.

I left a `TODO:` above about this that we can discuss.

pintc/src/predicate/optimize.rs
mohammadfawaz: This makes sense to me!

pintc/src/error/handler.rs
```diff
@@ -12,10 +12,10 @@ pub struct Handler {
 /// Contains the actual data for `Handler`.
 /// Modelled this way to afford an API using interior mutability.
 #[derive(Default, Debug)]
-struct HandlerInner {
-    /// The sink through which errors will be emitted.
-    errors: Vec<Error>,
-    // TODO: add warnings here
+pub struct HandlerInner {
```
mohammadfawaz, line 15: So this used to be private on purpose. I don't think we should make it public. Is this necessary?

pintc/src/error/handler.rs
```diff
@@ -12,10 +12,10 @@ pub struct Handler {
...
+    /// The sink through which errors and warnings will be emitted.
+    pub errors: Vec<Error>,
+    pub warnings: Vec<Warning>,
```
mohammadfawaz, line 18: Same with these. We shouldn't have access to `errors` and `warnings` directly like this. We have the API below to access and manipulate these.

pintc/src/error/handler.rs
```diff
@@ -55,17 +80,18 @@ impl Handler {
         }
     }
 
-    /// Extract all the warnings and errors from this handler.
-    pub fn consume(self) -> Vec<Error> {
-        let inner = self.inner.into_inner();
-        inner.errors
+    pub fn consume(self) -> HandlerInner {
```
mohammadfawaz, line 83: `consume` should probably look like this now:
```rust
    pub fn consume(self) -> (Vec<CompileError>, Vec<CompileWarning>) {
        let inner = self.inner.into_inner();
        (inner.errors, inner.warnings)
    }
```

pintc/src/error/handler.rs
```diff
@@ -55,17 +80,18 @@ impl Handler {
...
+        self.inner.into_inner()
     }
 
     pub fn append(&self, other: Handler) {
-        let errors = other.consume();
-        for err in errors {
+        let other_inner = other.consume();
```
mohammadfawaz, line 88: Similarly, `I would implement `append` as follows (after you change `consume`):
```rust
    pub fn append(&self, other: Handler) {
        let (errors, warnings) = other.consume();
        for warn in warnings {
            self.emit_warn(warn);
        }
        for err in errors {
            self.emit_err(err);
        }
    }
```

pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -1,22 +1,32 @@
...
 
 /// Remove all unused States in their respective predicates.
-pub(crate) fn dead_state_elimination(contract: &mut Contract) {
+pub(crate) fn dead_state_elimination(contract: &mut Contract, handler: &Handler) {
```
mohammadfawaz, line 29: We generally like to put the `handler` first in the list of arguments (or after `self` if we have it).

pintc/src/warning.rs
```diff
@@ -0,0 +1,211 @@
...
+pub struct Warnings(pub Vec<Warning>);
+
+impl Display for Warnings {
+    fn fmt(&self, f: &mut Formatter) -> Result {
+        write!(
+            f,
+            "{}",
+            self.0
+                .iter()
+                .map(|err| err.display_raw())
```
mohammadfawaz, line 99: `warning` instead of `err` just for clarity.

pintc/tests/optimizations/dead_state_elimination.pnt
mohammadfawaz: Let's remove this warning for now and address it in a separate issue. We first need to make sure that there is no `mut` anywhere in the `state` initializer before we emit a warning.

pintc/src/error/handler.rs
```diff
@@ -25,19 +25,44 @@ impl Handler {
         ErrorEmitted { _priv: () }
     }
 
+    /// Emit the warning `warn`.
+    pub fn emit_warn(&self, warning: Warning) -> WarningEmitted {
+        self.inner.borrow_mut().warnings.push(warning);
+        WarningEmitted { _priv: () }
+    }
+
```
mohammadfawaz, line 33: I doubt `WarningEmitted` is going to be useful.. I suggest we remove it for now.

pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -1,22 +1,32 @@
...
+    handler: &Handler,
+) -> Result<(), WarningEmitted> {
+    dead_state_elimination(contract, handler);
+    dead_constraint_elimination(contract, handler);
+
+    if handler.has_warnings() {
+        Err(handler.continue_with_warnings())
+    } else {
+        Ok(())
+    }
```
mohammadfawaz, line 25: Yeah this is likely not needed. Let's take it out for now.

mohammadfawaz (changes requested):
This is a great start! First wave of comments :) Address them and I'll do the second wave after.

lethalgem:
> This is a great start! First wave of comments :) Address them and I'll do the second wave after.

Wave one down ✅

pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -66,14 +69,25 @@ pub(crate) fn dead_constraint_elimination(contract: &mut Contract) {
...
+                            handler.emit_warn(Warning::AlwaysFalseConstraint {
+                                span: constraint.span.clone(),
+                            });
+                        } else {
+                            handler.emit_warn(Warning::TrivialConstraint {
+                                span: constraint.span.clone(),
+                            });
+                        }
+
                         Some((i, b))
```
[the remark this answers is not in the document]
lethalgem: Nope! Good catch

pint-pkg/src/build.rs
```diff
@@ -376,6 +377,12 @@ fn build_pkg(
                 }
             };
 
+            // Print any warnings
+            if handler.has_warnings() {
+                let (_, warnings) = handler.consume();
+                pintc::warning::print_warnings(&pintc::warning::Warnings(warnings));
+            }
+
```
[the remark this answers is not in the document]
lethalgem: Just tested and I can confirm that we are not emitting warnings twice. We won't get to this stage of printing if we have any errors before this.

lethalgem:
@mohammadfawaz wave 2 down ✅

mohammadfawaz (approved):

mohammadfawaz (approved):

PR #870, opened 2024-09-10
Upgrade to latest access opcodes

mohammadfawaz:
After https://github.com/essential-contributions/essential-base
- Each state var now gets a single slot. 
- Each pub var now gets a single slot. This has implications on abi gen which I dealt with.
- Storage keys are still assigned as before but, if the storage access needs more than one key, we consolidate the returned data into a single slot.
- This generally simplifies asm gen quite a bit and makes use of the new access opcodes features.


mohammadfawaz:
@otrho A lot of the `StateMemory` stuff in asm gen is going to change again in https://github.com/essential-contributions/pint/pull/877. This just how things progressed while working on arbitrary state programs.

pintc/tests/test.pnt
```diff
@@ -1,3 +0,0 @@
-predicate test {
-    constraint 0 < true ? 11 : 22;
-}
```
otrho, line 3: How did this sneak in?  Heh.

otrho (approved):

PR #872, opened 2024-09-11
Add front-end support for unions and `match` expressions and declarations.

otrho:
This PR has most of the changes needed for `union`s and `match`.

- It covers the front- and middle- ends; parser, type-checker and lowering.
- It has _some_ ASM gen support, currently untested.
- It has all the E2E testing to cover all possible errors and general use.

It does not have ABI generation nor validation testing.  There are no documentation changes yet either.

I have been working on this for several weeks and have made pretty much all of the design and implementation decisions unilaterally, and so many of these choices need to be reviewed, particularly before we document it.

To get an example of how `union`s and both `match` expressions and `match` declarations work presently you can take a look at the E2E tests, especially `unions/simple.pnt`.

https://github.com/essential-contributions/pint/blob/948b540067db248f0296942a658815e1c577b36e/pintc/tests/unions/simple.pnt

I think perhaps if it's OK we could merge this PR and add issues for all the remaining work.  Contracts which use `union`s won't compile (due to missing ABI) but since we're still pre-release I'd like to get it in since it'd getting too big and merge conflicts are an issue.

The main reason I didn't implement ABI output was I made an attempt and frankly got quite lost.  I'll need some help to understand what is required there.

There are still a couple of design issues to work out too.  Especially around storage -- do we allow `union`s in storage at all?  Also, do we allow match bindings to shadow each other?  Perhaps, since they're local in scope.

### DIFF
--- a/pintc/src/asm_gen/asm_builder.rs
+++ b/pintc/src/asm_gen/asm_builder.rs
@@ -277,7 +277,7 @@ impl AsmBuilder<'_> {
             }
         }
 
-        match &expr.get(contract) {
+        match expr.get(contract) {
             Expr::Immediate { value, .. } => {
                 compile_immediate(asm, value);
                 Ok(Location::Value)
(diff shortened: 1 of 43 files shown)


otrho:
While working on this I decided #874 and #875 are needed.  Unions make these issues worse.

mohammadfawaz (approved):
This all seems reasonable to me

PR #873, opened 2024-09-13
Lower nested `enums`

### LINKED ISSUE #648: Compiler internal error during flattening in simple contract with enum: `custom type present in final intent expr_types slotmap`
While working on #630, I ran into an error in my simple test contract and noticed that it still occurred without any dependencies.

Here's the pint contract I'm compiling using `master` version of `pintc`:

```
enum Animal = Cat | Dog;

type Person = {
    age: int,
    pet: Animal,
};

var bob = {
    age: 42,
    pet: Animal::Cat,
};

constraint bob.pet != Animal::Dog;
solve satisfy;
```

And here's the error:

```
$ pintc contract.pnt
Error: compiler internal error: custom type present in final intent expr_types slotmap
   ╭─[contract.pnt:1:1]
   │
 1 │ enum Animal = Cat | Dog;
   │ ───────────┬───────────
   │            ╰───────────── custom type present in final intent expr_types slotmap
───╯
Error: compiler internal error: custom type present in final intent expr_types slotmap
   ╭─[contract.pnt:1:1]
   │
 1 │ enum Animal = Cat | Dog;
   │ ───────────┬───────────
   │            ╰───────────── custom type present in final intent expr_types slotmap
───╯
Error: could not compile `contract.pnt` due to 2 previous errors
```

lethalgem:
Closes #648 

Wip*

Currently only lowers nested enums for vars and exprs. It _does not_ handle adding the proper bound constraints for the enum variants


pintc/src/types.rs
```diff
@@ -175,6 +175,56 @@ impl Type {
...
+            Type::Map { ty_from, ty_to, .. } => {
+                ty_from.has_nested_enum() || ty_to.has_nested_enum()
+            }
+
+            Type::Vector { ty, .. } => ty.has_nested_enum(),
+
+            Type::Error(_) | Type::Unknown(_) | Type::Any(_) | Type::Primitive { .. } => false,
+        }
+    }
+
```
lethalgem, line 227: Here I've used the same logic to either
1. replace a nested `enum ty` with the `int type` -- for use with `exprs` and `vars`
2. report whether or not there is a nested `enum ty` -- for reporting an error when `enums` are present in `states`

Should these be combined or remain separate?

lethalgem:
>Currently only lowers nested enums for vars and exprs. It does not handle adding the proper bound constraints for the enum variants

Did we want to address this in this PR? or should I just open an issue for it?

otrho:
What is this actually about?  I'm a bit confused.  Usually when something is 'nested' is found within itself.  When I see 'nested enum' I'm thinking 'an enum in an enum' which is impossible.

So we already have enum lowering, converting enums to their ints.  Is this improving that pass?

lethalgem:
> What is this actually about? I'm a bit confused. Usually when something is 'nested' is found within itself. When I see 'nested enum' I'm thinking 'an enum in an enum' which is impossible.
> 
> So we already have enum lowering, converting enums to their ints. Is this improving that pass?

'Nested' isn't the right term here then. This improves the pass by converting `enums` that are inside of compound types like `tuples` into `ints`.

ex. it removes the error we saw here
```pnt
enum Animal = Cat | Dog;

type Person = {
    age: int,
    pet: Animal,
};

predicate Foo {
    var bob = {
        age: 42,
        pet: Animal::Cat,
    };

    constraint bob.pet != Animal::Dog;
}

// --print-flat
predicate ::Foo {
    var ::bob: {child: {age: int, pet: ::Animal}, pet: ::Animal};
    var ::friends_pets: ::Animal[3];
    constraint (::bob == {child: {age: 23, pet: 1}, pet: 0});
    constraint (::friends_pets == [1, 0, 0]);
    constraint (::bob.pet != 1);
}

Error: compiler internal error: custom type present in final predicate expr_types slotmap
   ╭─[test.pnt:1:1]
   │
 1 │ enum Animal = Cat | Dog;
   │ ───────────┬───────────  
   │            ╰───────────── custom type present in final predicate expr_types slotmap
───╯
```

`var ::bob: {child: {age: int, pet: ::Animal}, pet: ::Animal};` still contains Animal, when it should be `int`. This was happening because we were checking if the `tuple` itself was an `enum` and not if an enum was used for any of the fields.

otrho:
> ex. it removes the error we saw here
> 
> `var ::bob: {child: {age: int, pet: ::Animal}, pet: ::Animal};` still contains Animal, when it should be `int`. This was happening because we were checking if the `tuple` itself was an `enum` and not if an `enum` was used for any of the fields.

Oh, wow, this is surprising.  I thought we'd already fixed all that.  Fair enough.

I also often see the flattened output in tests with certain things like alias and enum decls still there when they aren't used at all by the contract -- because they've been flattened -- and think we should just remove those declarations so they're not even printed.  Just to allay any confusion around them.

pintc/src/predicate/exprs.rs
```diff
@@ -97,12 +97,18 @@ impl ExprKey {
...
 
-    /// Returns the type of key `self` given an `Predicate`. Panics if the type can't be
+    /// Returns the type of key `self` given a `Contract`. Panics if the type can't be
     /// found in the `expr_types` map.
     pub fn get_ty<'a>(&'a self, contract: &'a Contract) -> &Type {
         contract.exprs.expr_types.get(*self).unwrap()
     }
 
+    /// Returns the mutable type of key `self` given a `mu Contract`. Panics if the type can't be
+    /// found in the `expr_types` map.
```
mohammadfawaz, line 107:
```suggestion
    /// Returns a mutable reference to the type of key `self` given a `Contract`. Panics if the type can't be
    /// found in the `expr_types` map.
```

pintc/src/predicate/vars.rs
```diff
@@ -93,6 +93,12 @@ impl VarKey {
         pred.vars.var_types.get(*self).unwrap()
     }
 
+    /// Returns the mutable type of key `self` given a `mut Predicate`. Panics if the type can't be
+    /// found in the `var_types` map.
```
mohammadfawaz, line 97:
```suggestion
    /// Returns a mutable reference to the type of key `self` given a `Predicate`. Panics if the type can't be
    /// found in the `var_types` map.
```

mohammadfawaz (commented):
Yeah this is good. Please open an issue about adding the missing constraints.

mohammadfawaz (approved):

pintc/tests/unions/simple.pnt
```diff
@@ -109,8 +109,8 @@ predicate test {
 // predicate ::test {
 //     var ::x: ::thing;
 //     var ::d: int;
-//     var ::no_base_addr: ::maybe_addr;
-//     var ::a_base_addr: ::maybe_addr;
+//     var ::no_base_addr: int;
+//     var ::a_base_addr: int;
```
otrho, line 113: This is not right -- two `var`s which were unions are now integers.  I'll look to fix it in my current branch.

PR #877, opened 2024-09-16
Arbitrary state programs

### LINKED ISSUE #773: Abi gen pub vars
Currently the abi gen handles pub vars like state mutations but I think the current implementation of pub vars expects that all pub vars are set.
Does this work?
```rust
pub var a: int;

constraint a == nil || a == 1;
```
I think we really need to pick a direction with pub vars. They are either like storage and can eventually support things like `map` and it's possible for them to be nil like storage (and we can check for nil).
Or they are the same as vars and must all be set.

I think the first option is much more powerful and fits the original intention for pub vars but currently it's very confusing being in this middle ground.

Users won't realize that they have to set all the pub vars because it's not a struct like vars are.

### LINKED ISSUE #860: Support remaining arbitrary types in storage vectors
Currently we only support `bool`, `int`, and `b256`. Support everything else including maps.

mohammadfawaz:
Closes #860 
Closes #773

Enables arbitrary state programs, i.e., arbitrary expressions as `state` var initializers. Most things should work. I think there are some exceptions such as generators, but those are not properly supported yet anyways.

List of changes:
* parser now accepts storage accesses as expressions (i.e. can be used anywhere), but the middleend will stop illegal ones later. So, storage access can still only be used in state var initializers.
* Lots more changes in asm gen to how state slots are assigned and to support arbitrary state programs. Most things are documented well I'd say.
* `__state_len` is now `__size_of` and can be used anywhere. It will just yield the size of the value which can be 0 if the value can be `nil`. This is only the case for paths to state vars AND storage accesses.


pintc/src/predicate/transform/lower.rs
```diff
@@ -554,7 +554,7 @@ pub(crate) fn lower_array_ranges(
         if let Some((storage_vars, _)) = storage {
             array_range_expr_keys.extend(storage_vars.iter().filter_map(
                 |StorageVar { ty, .. }| {
-                    dbg!(ty.get_array_range_expr());
```
otrho, line 557: I literally found this about 10 minutes ago myself!  Another whoops.

otrho (approved):

PR #884, opened 2024-09-17
*WIP: Optimize `Selects`

### LINKED ISSUE #883: Optimize selects
E.g.: the expression
```
true ? 1 : 2;
```
can be changed to just `1`.

So basically, for every select expression, const evaluate the condition, and if the result is `true`, then just replace the select expression with the first branch. Otherwise, replace it with the second branch.

lethalgem:
Closes #883


lethalgem:
I can't remember what was done in the past for pintc/src/asm_gen/tests.rs

Do we modify the test to account for the optimization pass? Or should the test input be modified to prevent optimizing out anything?

We have a flag for the tests that specifies that we don't skip the optimization phase, so I know we specifically want the phase to run.

pintc/tests/optimizations/dead_select_elimination.pnt
```diff
@@ -0,0 +1,92 @@
+predicate test {
+    // select exprs that should be optimized out
+    var a = true ? 1 : 2;
+    var b = false ? 1 : 2;
+    var c = (1 + 2 == 3) ? 1 : 2;
+    var d = (1 < 2) ? 1 : 2;
+
+    // should this be removed? The evaluator can't handle this right now
+    var e = 2 < 1;
+    var f = e ? 1 : 2;
```
lethalgem, line 10: Comment covers it here. We faced a similar issue with #834 where we currently don't remove `constraints` that have `vars` in the keys for the `constraint`.

It's currently just a limitation of the `evaluator` so maybe that should be opened as a new issue? Or maybe we don't want to remove things like this and `evaluator` is working as intended?

mohammadfawaz: Yeah that's a more "advanced" evaluation.. let's not worry about it just yet, though it wouldn't be too difficult to do.

mohammadfawaz:
> Do we modify the test to account for the optimization pass? 

Yes. We should make sure to change the tests to not trigger dead selection elimination, while keeping them as simple as possible (we don't want that file to blow up!)

pintc/src/warning.rs
```diff
@@ -66,10 +66,19 @@ impl ReportableWarning for Warning {
...
             }
 
+
+
+
+
+
+
+
+
```
mohammadfawaz, line 81:
```suggestion
```

mohammadfawaz (commented):
This is pretty good!

pintc/src/asm_gen/tests.rs
```diff
@@ -179,10 +179,7 @@ fn select() {
                   Stack(Push(0))
                   Stack(Push(1))
                   Access(DecisionVarRange)
-                  Stack(Push(69))
```
mohammadfawaz, line 182: A reminder to change those tests to avoid selects from getting optimized away.

lethalgem: 🤦, fixed and didn't push

mohammadfawaz (approved):
Awesome :shipit:

Issue #874, opened 2024-09-14
Improve `Type::Custom` (maybe with a `Type` refactor).


otrho:
After we add unions there are 3 different 'custom' types -- aliases, enums and unions.  They're all distinct and shouldn't really be handled by the same code.

Ideally we remove custom types before type checking.  We do already have `lower_custom_types()` as the first step in the type checker and maybe it's doing enough.  We need to ensure there are no custom types left in the contract after this step though to simplify checking and the rest of `types.rs`.  Unions are still assuming they can be `Type::Custom` which is not a great design.

Even more ideal is for `Type::Custom` and `Type::Unknown` and even `Type::Error` to not even be valid variants after type checking.  We would have a pre-checked type for `Type` and post-checked type (or type-checked `Type`).  The former could just be a wrapper around `Type` which has `Error`, `Unknown`, `Custom` and `Concrete(Type)` as variants.  But this would require `Contract` to have its `Type` parameterised and for the type checker to take a `Contract<Wrapped<Type>>` and produce a `Contract<Type>`.  This sort of a refactor might be a bit ambitous at this stage.  I *did* say 'ideally'...

PR #888, opened 2024-09-19
Further adventures in enum lowering.

otrho:
#873 was a bit aggressive and converted some union types into integers erroneously.

- move the nested enum code from `types.rs` into `lower.rs` since that's the only place it'll get called.
- make the nested checking more `Type::Custom` aware -- they're not necessarily enums.
- add a couple more instances to tests; `const`s and `interface`s.

This change adds to the argument for #874 to be addressed.

### DIFF
--- a/pintc/src/asm_gen/asm_builder.rs
+++ b/pintc/src/asm_gen/asm_builder.rs
@@ -253,7 +253,7 @@ impl AsmBuilder<'_> {
     ) -> Result<Location, ErrorEmitted> {
         fn compile_immediate(asm: &mut Asm, imm: &Immediate) {
             match imm {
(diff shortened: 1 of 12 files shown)


otrho:
While doing this I've taken a digression into fixing this:

```
enum x = a | b;

const y: { x, x } = { x::a, x::b };

predicate test {
}
```

```
Error: const initialization type error
   ╭─[test.pnt:3:21]
   │
 3 │ const y: { x, x } = { x::a, x::b };
   │          ────┬───   ───────┬──────
   │              ╰────────────────────── expecting type `{::x, ::x}`
   │                            │
   │                            ╰──────── const initializer has unexpected type `{int, int}`
───╯
```

mohammadfawaz (approved):

PR #889, opened 2024-09-19
Keep track of `span` when `lowering ifs`

### LINKED ISSUE #885: Warning without a span
The following emits a warning without a span:
```rust
predicate Foo { 
    var b: bool;
    if true {
        constraint 1 < 2;
    } 
}
```
Warning:
```
Warning: constraint is always `true`
   ╭─[:1:1]
   │
 1 │ <none>
   │ │
   │ ╰─ this constraint always evaluates to `true`
   │
   │ Help: if this is intentional, consider removing this constraint
```
That's because during flattening, we first convert the `if` statement to:
```rust
    constraint (!true || (1 < 2));
```
and I guess the span get lost in the process.

lethalgem:
Closes #885

### DIFF
--- a/pintc/src/predicate/transform/lower.rs
+++ b/pintc/src/predicate/transform/lower.rs
@@ -1070,7 +1070,7 @@ fn convert_if(
         Expr::UnaryOp {
             op: UnaryOp::Not,
             expr: *condition,
-            span: condition.get(&contract).span().clone(),
+            span: condition.get(contract).span().clone(),
         },
         Type::Primitive {
             kind: PrimitiveKind::Bool,
--- a/pintc/src/predicate/transform/lower.rs
+++ b/pintc/src/predicate/transform/lower.rs
@@ -1037,15 +1037,15 @@ pub(crate) fn lower_ifs(contract: &mut Contract) {
 
         // Ideally we'd refactor this to not require cloning the IfDecl.
         for if_decl in &contract.preds[pred_key].if_decls.clone() {
-            all_exprs.extend(&convert_if(contract, pred_key, if_decl));
+            all_exprs.extend(convert_if(contract, pred_key, if_decl));
         }
 
         let mut_pred = contract.preds.get_mut(pred_key).unwrap();
 
-        for expr in all_exprs {
+        for (expr, original_expr_span) in all_exprs {
             mut_pred.constraints.push(ConstraintDecl {
                 expr,
-                span: empty_span(),
+                span: original_expr_span,
             });
         }
 
@@ -1065,12 +1065,12 @@ fn convert_if(
         else_block,
         ..
     }: &IfDecl,
-) -> Vec<ExprKey> {
+) -> Vec<(ExprKey, Span)> {
     let condition_inverse = contract.exprs.insert(
         Expr::UnaryOp {
             op: UnaryOp::Not,
             expr: *condition,
-            span: empty_span(),
+            span: condition.get(&contract).span().clone(),
         },
         Type::Primitive {
             kind: PrimitiveKind::Bool,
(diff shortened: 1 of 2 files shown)


pintc/src/predicate/transform/lower.rs
```diff
@@ -1037,15 +1037,15 @@ pub(crate) fn lower_ifs(contract: &mut Contract) {
 
         // Ideally we'd refactor this to not require cloning the IfDecl.
         for if_decl in &contract.preds[pred_key].if_decls.clone() {
-            all_exprs.extend(&convert_if(contract, pred_key, if_decl));
+            all_exprs.extend(convert_if(contract, pred_key, if_decl));
```
lethalgem, line 1040: From my limited understanding this is still functionally the same because of how `.extends()` handles a return type of `Vec<ExprKey>` and `Vec<(ExprKey, Span)>`

lethalgem:
I extended the test in pintc/tests/optimizations/false_constraint_simplification.pnt to report the proper span for the `warning`. It made sense to me because this test is testing the section of the `optimizer` where the `warning` originates.

We don't have a specific directory to test for `errors`. So I feel like we shouldn't have one specifically for `warnings` either.

lethalgem:
@mohammadfawaz When we last chatted, we were both wondering what the behavior of the below was:

```rust
if false {
    constraint 1 > 2;
}

Warning: constraint is always `true`
   ╭─[test.pnt:3:9]
   │
 3 │         constraint 1 > 2;
   │         ───┬────
   │                 ╰───────── this constraint always evaluates to `true`
   │ 
   │ Help: if this is intentional, consider removing this constraint
───╯
```
Not only does this return a warning, it's the opposite of what the user would expect. This likely has to do with the `condition_inverse` in `lower_ifs()`.  So the warning is for the newly made constraint evaluating to `true`, but the original `expr` doesn't reflect that.

otrho:
> //     constraint (!false || (1 > 2));

So it looks like testing for redundant constraints should be done before `if` lowering.  OTOH, `match`es would also still exist then and we wouldn't be able to check for redundant constraint within.

pintc/src/predicate/transform/lower.rs
otrho: Did you need to remove the `&` reference for it to compile?  `extend()` takes any type which implements `IntoIterator` and the result of `convert_if()` is a temporary rvalue so I'm surprised it makes any difference.

Taking the `&` returns a slice which would then be owned by the iterator and just passing the value would mean the `Vec` itself is owned by the iterator but either should work.  The only difference is if you wanted to use the result of `convert_if()` later on, since it's moved into `extend()`.

otrho (approved):

pintc/src/predicate/transform/lower.rs
lethalgem: I did need to remove the `&`

When I keep the `&`, rust-analyzer says I need to dereference `expr` when pushing `ConstraintDecl`

```rust
for if_decl in &contract.preds[pred_key].if_decls.clone() {
    all_exprs.extend(&convert_if(contract, pred_key, if_decl));
}

let mut_pred = contract.preds.get_mut(pred_key).unwrap();

for (expr, original_expr_span) in all_exprs {
    mut_pred.constraints.push(ConstraintDecl {
        expr: *expr,
        span: original_expr_span,
    });
}
```

Which makes sense to me. But once I do that, I get 
>temporary value dropped while borrowed
>consider using a `let` binding to create a longer lived value

for `&convert_if(contract, pred_key, if_decl)` which I believe makes the use of `&` pointless here because I would need to hold on to the values anyway?


FWIW when I use `&`, `all_exprs` is of type `Vec<&(ExprKey, Span)>`, and when I don't use `&`, `all_exprs` is of type `Vec<(ExprKey, Span)>`. Which reflects what `all_exprs` used to be (`Vec<ExprKey>`) before `convert_if` returned a `tuple`.

lethalgem: Happy to change it of course. Just let me know what you think and I can make a quick PR to modify :D

PR #895, opened 2024-09-25
Split `type_check.rs` into smaller modules.

otrho:
I have been updating `type_check.rs` recently and noticed the file size was getting over 3000 lines.  So I thought 'we should split that up, it's too big'.

This is the traditional idea; I was told that a source file should never go above 1000 lines and refactoring it into smaller modules is the answer.  But this was in the olden days, when a) compilers (and disks and/or filesystems) were slower, and b) before we had LSP for every language under the sun, allowing simple access to any parts of a codebase with fuzzy finding.

I went and split `type_check.rs` anyway.  But I'm thinking this actually won't make much difference to how I work anyway.  As above, I'd use LSP to find what I need and which file or module it's in doesn't really matter.  It's good to have similar code together so you're not jumping around when using prior art as inspiration, but a 3000 line file perhaps isn't the end of the world any more..?

So IDK, perhaps this PR isn't worth merging?  I'd be interested to hear opinions.  @mohammadfawaz @lethalgem 

The changes are purely at the method level of granularity, nothing inside a method has changed, just which file/module it's in.  If you have changes to `type_check.rs` in a working branch then it might be a pain to merge since Git can't do it for new files.


mohammadfawaz (approved):
Yeah this makes sense to me.

lethalgem:
I don't have any strongly held opinions here. Mostly looking to ya'll to learn and gain experience around when files should be split up.

to preface: I've heard this about both files **_and_** functions

In the past I've heard that files shouldn't be bigger than even a few hundred lines (and functions less than 20 lines, if you can believe it). That the bigger the file was, the more it was a hint to abstract away granular pieces so that the reader could understand the high level logic faster. I worked with a project like this. Nearly every file was kept to under 100 lines. I found it to be a minor annoyance. Just constantly clicking around different files to find where I needed to make changes.

Recently I've heard the opposite. That this brings in more indirection and makes it harder to find what you're looking for. Having related methods be in close proximity makes it easier to scan and end up where you need to be, faster.

Looking at the changes, I like splitting up this file. I almost never find myself scrolling through the entire `type_check` file. I either search, or cmd+click to follow the definitions of functions until I end up where I need to make changes. Plus, the files are well named and make it easy enough to navigate without the aid of the editor.

otrho:
> In the past I've heard that files shouldn't be bigger than even a few hundred lines (and functions less than 20 lines, if you can believe it).
>
> Recently I've heard the opposite. That this brings in more indirection and makes it harder to find what you're looking for. Having related methods be in close proximity makes it easier to scan and end up where you need to be, faster.

Yeah, Java introduced the idea of a single class per file, and they have to have the same name, and many C++ projects I worked on were the same.  No matter how small your class was it had to go in its own file.  And if you couldn't see an entire function on your screen at once then it needed to be split up.  Nice ideas but I think they're from the before-times, when editors and IDEs were shit.

In the very early days of my career we had a codebase which put all function names at the start of the line, so you could find them using a `^name` regex search:
```
int
count_frogs(struct pond_t* pond)
{
    ...
}
```

PR #898, opened 2024-09-26
Optimize simple exprs

lethalgem:
_Draft_
* `asm_gen` tests are currently failing. I just wanted to get the PR up so @mohammadfawaz could take a look at it

This PR optimizes `BinaryOp exprs` by folding any const `exprs`

Ex. `var c = (1 + 2) - 3 + 4 - (2 - 5);` becomes `constraint (::c == 7);`


pintc/src/predicate/optimize/expr_simplification.rs
```diff
@@ -0,0 +1,48 @@
...
+        let exprs_to_check = contract
+            .exprs(pred_key)
+            .filter_map(|expr_key| {
+                if let Some(Expr::BinaryOp { span, .. }) = expr_key.try_get(contract) {
+                    Some((expr_key, span.clone()))
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>();
```
lethalgem, line 33: @mohammadfawaz 

What other `exprs` would be simplified by this pass? Looking at all of our `exprs`, all I could think of were simple `BinaryOp`s

lethalgem:
@mohammadfawaz 

The original goal of this PR was:
```
constraint x + 1 + 2 == 4;
should become
constraint x + 3 == 4
```

_technically_ this is the case
```
constraint (::c == ((((1 + 2) - 3) + 4) - (2 - 5)));
becomes
constraint (::c == 7);
```

I don't think this was the spirit of the issue though. I _think_ we wanted
```
Predicate Test {
    var x: int;
    constraint x + 1 + 2 == 4; 
}

to result in 

constraint ((::x + 3) == 4)
```

but unfortunately the `flattened` version of the constraint is `constraint (((::x + 1) + 2) == 4);` So when running over `exprs()` and evaluating `immediates`, there is nothing to simplify. `(x + 1)` would not be changed and then we're left with the original `(x + 1) + 2` expr.

I spent a good while looking at how I could simplify consts that were next to each other but not in the same ast node, and didn't come up with a good solution.

Let me know if you have any ideas or if I'm overthinking this.

mohammadfawaz:
> I spent a good while looking at how I could simplify consts that were next to each other but not in the same ast node, and didn't come up with a good solution.

This is actually not that simple.. don't worry about it for now. We'll open a separate issue to handle it.

pintc/src/predicate/optimize/expr_simplification.rs
lethalgem: Addressed

pintc/src/predicate/optimize/const_folding.rs
```diff
@@ -0,0 +1,52 @@
...
+
+                let simplified_expr = Expr::Immediate {
+                    value: imm.clone(),
+                    span: expr_key.get(contract).span().clone(),
+                };
+                replace_map.insert(
+                    expr_key,
+                    (simplified_expr, expr_key.get_ty(contract).clone()),
+                );
+            }
```
lethalgem, line 35: @mohammadfawaz Is there a way to tell if the new `imm` is a fold of the original `expr`? I've been struggling to think of a way to accomplish this check.

If we don't have a check here, then we replace every `expr` even if it doesn't need it. So the `replace_map` essentially duplicates everything in `contract.exprs`. Then it goes through the work of replacing every single `expr` even if we don't need to make a single replacement.

ex. 

```
predicate Test {
    var a = 1 + 2;
}

replacing --- 
old expr: 2
with
new expr: 2
--- 
replacing --- 
old expr: {0}
with
new expr: {0}
--- 
replacing --- 
old expr: 1
with
new expr: 1
--- 
replacing --- 
old expr: 0
with
new expr: 0
--- 
replacing --- 
old expr: (1 + 2)
with
new expr: 3
--- 
```

only the last `expr` above needs to be replaced.

A really dirty way to accomplish this would be to string check 
```
contract.with_ctrct(old_expr_key), != contract.with_ctrct(simplified_expr_key),
```

which I don't believe we'd be alright doing.

lethalgem: Added. Seems to work great! Thanks @otrho

Issue #879, opened 2024-09-17
Unions: Minor fixes and/or updates.


otrho:
#872 introduced a WIP implementation of unions.  There are a few items which may or may not be issues in need of attention.

- [x] Unions are initially `Type::Custom`, as are aliases.  There may be issues where aliases which declare unions, or unions containing aliases, or unions containing other unions, or more complicated forms of nesting may not be lowered correctly.  Also see #874.
- [x] The equality tests (with the `==` op) between two union variant values should be possible and compile correctly.  E.g., `val == union_t::variant(11)`.
- [x] Shadowing needs to be formalised.  See below.
- [x] Compile time evaluation of internal union expressions (`UnionTagIs` and `UnionValue`) should also work.  After lowering, these expressions appear within `if` or `?:` expressions and if the union values are immediates they should be able to be evaluated.

Shadowing of symbols is generally disallowed in Pint, but that is only between the global (`const`) scope and the predicate scope(s), or within those two scopes.  Match bindings are of a more limited nature and it may be nice to allow Rust style 'unwrapping', e.g.,
```
match shape {
    option::some(shape) => ... // `shape` is now 'unwrapped'.
    option::none => ...
}
```
Or perhaps this isn't OK and we should ban shadowing of any kind throughout Pint.

PR #905, opened 2024-10-01
Add compile time evaluation for union literals.

otrho:
Another dot-point from #879.


otrho:
It exaggerates the poor design of unions + `Type::Custom` + `Expr::Path` some more too.  Part of the solution for #874 should be to use a `slotmap` and a `UnionDeclKey` or similar, so looking up union details like the variant names and types doesn't use string compares of (sub-) paths.  And probably to remove `enum` altogether, it would seem.

mohammadfawaz (approved):

PR #906, opened 2024-10-02
Remove enums and strictly replace them with unions

### LINKED ISSUE #858: Support Enums in storage including storage maps and storage vectors.

mohammadfawaz:
* Removes enums completely
* Replaces all places where enums were used (in code and tests) with unions that behave like enum.. I'm calling them "enumeration unions".
* Allow unions in storage. We need to address the ABI next.

closes #858


pintc/src/expr/evaluate.rs
```diff
@@ -366,13 +391,34 @@ impl Evaluator {
...
+                    // enumeration
+                    Imm::UnionVariant {
+                        tag_num,
+                        ref ty_path,
+                        ..
+                    } => {
+                        let is_enumeration_union = Type::Union {
+                            path: ty_path.clone(),
+                            span: empty_span(),
+                        }
```
mohammadfawaz, line 404: Creating the type like this from `ty_path` is not great and I think we're doing that somewhere else already so I did the same... but that should be improved.

otrho: Yep, I'm keen now to have the union decls referred to by `slotmap` key instead of the crappy string matching the `Custom` types (and similar) do now.

otrho (dismissed):

otrho (approved):

PR #911, opened 2024-10-07
Support `in` exprs in `evaluator`

### LINKED ISSUE #886: `in` expressions in `const` initializers
Low priority but the following should really work:
```rust
const x = 1 in 1..2;
```
and so is
```rust
const x = 1 in [1, 2];
```
Currently, we just get an internal error:
```console
Error: compiler internal error: unexpected expression during compile-time evaluation
```

lethalgem:
Closes #886

### DIFF
--- a/pintc/src/expr/evaluate.rs
+++ b/pintc/src/expr/evaluate.rs
@@ -517,6 +517,10 @@ impl Evaluator {
                             Ok(Imm::Bool(ub >= value && lb <= value))
                         }
 
+                        (Imm::Real(lb), Imm::Real(ub), Imm::Real(value)) => {
+                            Ok(Imm::Bool(ub >= value && lb <= value))
+                        }
+
                         _ => Err(handler.emit_err(Error::Compile {
                             error: CompileError::Internal {
                                 msg: "unexpected expression during compile-time evaluation \
--- a/pintc/src/expr/evaluate.rs
+++ b/pintc/src/expr/evaluate.rs
@@ -503,12 +503,30 @@ impl Evaluator {
                     }
                 }),
 
+            Expr::In {
+                value, collection, ..
+            } => {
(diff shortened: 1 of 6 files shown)


pintc/src/expr/evaluate.rs
```diff
@@ -503,12 +503,50 @@ impl Evaluator {
...
+                            Ok(Imm::Bool(ub >= value && lb <= value))
+                        }
+
+                        _ => Err(handler.emit_err(Error::Compile {
+                            error: CompileError::Internal {
+                                msg: "unexpected expression during compile-time evaluation \
+                            evaluation of in expr with range",
+                                span: empty_span(),
+                            },
+                        })),
```
lethalgem, line 526: My understanding is that we only support `ranges` with `ints` and that `reals` are experimental. If that is the case, then it should be fine to only evaluate `in exprs` structured as `int in int..int`?

Also I'll be updating the error messages, as I don't believe they should be internal. A user can still write `const f = true in 3..2;` without an error occurring before this stage.

FWIW, if it not a `const` then we type check the range. Ex.
```
predicate Test {
    var l = true in 1..2;
}

Error: value type and range type differ
    ╭─[test.pnt:18:21]
    │
 18 │     var l = true in 1..2;
    │                     ┬
    │                     ╰── range type mismatch; expecting `bool` type, found `int` type
────╯
```

mohammadfawaz: 1. We should support `real` regardless since we do handle it in other parts of the evaluator. Should be fairly straightofward. To test it, add a test under `pintc/tests/experimental/` and run with `cargo test --all-features`.
2. Yeah keep the internal error for now but open an issue to error out properly in the type checker.

mohammadfawaz (approved):
Looks reasonable to me!

PR #914, opened 2024-10-09
Verify that all non-storage types work for `pub vars`

### LINKED ISSUE #861: Verify that all non-storage types work for `pub var`s
I believe everything should work fine but we don't have sufficient testing.

lethalgem:
Staying in DRAFT:
- PR may not apply if we decide to remove `pub vars`. TBD
-----

Closes #861 

Note: 
~~Type Aliases for Unions currently don't work~~ _This is known and in progress_
```
union A = a(int) | b | c(bool) | d(b256) | e({int, int}) | f(int[2]);
type NewAlias = A;

predicate Simple {
    pub var a: NewAlias;

    constraint a == NewAlias::a(3);
}

Error: unknown union
    ╭─[test.pnt:24:21]
    │
 24 │     constraint a == NewAlias::a(3);
    │                     ─────┬─────  
    │                          ╰─────── union declaration for `::NewAlias::a` not found
────╯
```


mohammadfawaz:
No longer relevant since `pub var`s are gone

PR #916, opened 2024-10-15
Use a slotmap for union decls in `Contract`.

otrho:
This is an attempt to improve code quality around unions, and in general to try to lessen the amount of path/string compares we're using in the type system.  There's still more to do but this is a start and it took me ages since everything in this area is so fragile, so I'll push what I have here now.

It's supposed to be addressing some of #879 but I'm not sure it does really.  There may still be issues around nested aliases and unions.

This main difference here is that the `Type::Union` now has a key into a slotmap for the declaration rather than a name string.

Other changes:
- Nested new types are flattened completely early on.
- Custom types are lowered more aggressively before type checking.  Hopefully they can be practically removed post-type checking soon.
- `types::Path` is gone, we now just use a `String`.  This removes any illusions that `Path` was anything other than just a string.

### DIFF
--- a/pintc/src/predicate/analyse/type_check/custom_types.rs
+++ b/pintc/src/predicate/analyse/type_check/custom_types.rs
@@ -1,6 +1,6 @@
 use crate::{
     error::{CompileError, Error, ErrorEmitted, Handler},
-    predicate::{Const, Contract, Expr, Interface, InterfaceVar, StorageVar},
+    predicate::{Contract, UnionKey},
     span::{empty_span, Span},
     types::{NewTypeDecl, Type, UnionDecl},
 };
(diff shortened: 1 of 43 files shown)


pintc/src/predicate/analyse/type_check/custom_types.rs
```diff
@@ -113,118 +222,123 @@ impl Contract {
...
-            }
-        }
-
-        Ok(())
-    }
-
-    /// Lower every `Type::Custom` type in a contract to a `Type::Alias` type if the custom type
-    /// actually matches one of the new type declarations in the contract. All remaining custom
-    /// types after this method runs should refer to enums.
+    //fn lower_custom_types_in_newtypes(&mut self, handler: &Handler) -> Result<(), ErrorEmitted> {
```
mohammadfawaz, line 225: Are we going to need this in the future?

otrho: Ooh, I wasn't sure at the time, and then I forgot about it, and now all the tests pass, so I guess not.

PR #928, opened 2024-10-19
fixed error; need to upzip names and predicates after mapping

rybla:


pintc/src/asm_gen.rs
```diff
@@ -118,7 +118,9 @@ pub fn compile_contract(
                     })
                 })
         })
-        .collect::<Result<_, _>>()?;
+	.collect::<Result<Vec<(_, _)>, ErrorEmitted>>()?
+        .into_iter()
+        .unzip();
```
mohammadfawaz, line 123: While this makes sense and is probably clearer, Rust is actually able to auto-unzip as part of the `collect` operation. Meaning:
```
.collect::<Result<_, _>>()?;
```
is equivalent to
```
.collect::<Result<(Vec<_>, Vec<_>), _>>()?;
```
because of `let (names, predicates)` on line 105. Rust infers that it should collect into a tuple.

rybla: I didn't know about the impl of FromIterator that that unzips like that, neat

rybla: I was getting a compiler error here when I built locally. So I think the impl that accomplishes this was added in a later version of cargo than I was using, or something like that

mohammadfawaz: > I was getting a compiler error here when I built locally.

Ah that is quite odd. Yeah maybe older versions of Rust don't support this. FYI, we always build with the latest stable Rust build.

PR #935, opened 2024-10-21
Fold any `binaryop` that consists of a `boolean` or arithmetic identity

### LINKED ISSUE #926: Known conditionals should be shortcutted
When we know the condition of an `if statement`, we should optimize away any constraints for the unsatisfiable branch.

Ex.
```
predicate Foo {
    var x: int;
    var y: int;

    if true {
        constraint x == y;
    } else {
        constraint x != y;
    }
}

results in:

--print-flat

predicate ::Foo {
    var ::x: int;
    var ::y: int;
    constraint (!true || (::x == ::y));
    constraint (true || (::x != ::y));
}

--print-optimized

predicate ::Foo {
    var ::x: int;
    var ::y: int;
    constraint (false || (::x == ::y));
    constraint (true || (::x != ::y));
}

whereas it should be:

```

lethalgem:
Closes #926

`true || <expr>` is `true`
`true && <expr>` is `<expr>`
`false || <expr>` is `<expr>`
`false && <expr>` is `<false>`

x + 0 is x
x - 0 is x
x * 0 is 0
x * 1 is x
0 / x is 0
x / 1 is x
x % 1 is x


pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -53,12 +53,92 @@ pub(crate) fn dead_state_elimination(contract: &mut Contract) {
...
+/// true && <expr> is <expr>
+/// false || <expr> is <expr>
+/// false && <expr> is <false>
+///
 /// If any constraint evaluates to false, all constraints are removed and replaced with a single
 /// instance of `constraint false`
 pub(crate) fn dead_constraint_elimination(handler: &Handler, contract: &mut Contract) {
     let evaluator = Evaluator::new(contract);
 
     for pred_key in contract.preds.keys().collect::<Vec<_>>() {
```
lethalgem, line 67: I added the optimization to `dead_constraint_elimination` for two reasons.

1. Optimization passes are supposed to be independent of other passes. By simplifying in this pass, I didn't have to repeat any code for the cases of `true || <expr> is true` and `false && <expr> is <false>`. Otherwise I would have to make sure the new simplification took place before this pass or repeat the pass to handle `constraint false` and `constraint true`
2. It just felt like the natural home for this check

I don't know if it technically counts as dead code elimination or if it's more appropriately const folding? Or perhaps it is commonly referred to as something else altogether? If so, I'll move it or break it out into it's own pass 😄

pintc/tests/optimizations/const_folding.pnt
```diff
@@ -89,8 +89,7 @@ predicate Test {
 //     constraint (::d == true);
 //     constraint (::e == -23);
 //     constraint (::f == (::d ? false : true));
-//     constraint (true || (0 == ::y));
-//     constraint (false || (((::y == 1) || (::y == 4)) || (::y == 7)));
+//     constraint (((::y == 1) || (::y == 4)) || (::y == 7));
```
lethalgem, line 92: Unfortunately the new logic affects the `const_folding` test as well. I couldn't think of a way for it not to. When there is an `if statement` with a `const` condition, it's guaranteed to optimize away the `constraint true` and `constraint false` of the branches

pintc/src/predicate/optimize/dead_code_elimination.rs
mohammadfawaz: So one concern here is that we are overlooking Boolean exprs that could be optimized somewhere else. A simple example would be:
```rust
var y: int;
state x = true && y;
```
Your current implementation wouldn't catch that. What I'm saying is that, similar to const folding, we need to look at all expressions here, not just constraints. And technically these "Boolean identities" should be handled in constant folding.

That being said, I think we should expand the scope of this a little bit and include stuff arithmetic identities. Example:
```rust
x * 0 == 0
0 / x == 0
x * 1 == x
x / 1 == x
x + 0 == 0
x - 0 == 0
x % 1 == x
```
(there may be others.. think through them)

pintc/tests/optimizations/const_folding.pnt
mohammadfawaz: Yeah this is fine :)

pintc/src/predicate/optimize/dead_code_elimination.rs
lethalgem: Just moved boolean identity folding to `const_folding`. Will add arithmetic identities next

mohammadfawaz:
Note that you now need to move `var`s to predicate arguments and give them all types. If they have initalizers, then you turn them into a constraints.

pintc/src/predicate/optimize.rs
```diff
@@ -8,9 +8,9 @@ use crate::error::Handler;
...
 impl super::Contract {
     pub fn optimize(mut self, handler: &Handler) -> Self {
-        dead_code_elimination(handler, &mut self);
+        const_folding(&mut self); // TODO: Make passes independent. Right now we rely on const_folding before dead_constraint_elim
 
-        const_folding(&mut self);
+        dead_code_elimination(handler, &mut self);
 
         self
     }
```
lethalgem, line 16: @mohammadfawaz in the past I believe you and I have talked about how optimization passes need to operate independently and not depend on the order they're run in.

I'm not sure if this violates that rule or not. To get the optimal final result, `const_folding` needs to be done before `dead_code_elimination`. Specifically for the case where the constraint is `constraint true || x` -> `constraint true` -> removed as dead code.

If we do it the other way around, we're left with the `constraint true` because we checked if the `constraint true || x` should be removed before it was transformed to `constraint true`.

The reason I think it may not violate the rule is because both passes can be run at anytime without causing the other to error. Technically they remain independent of each other, you just don't get the best final result unless they're paired together in the correct order.

The only way I can think of having the order no matter would be to duplicate the logic in `dead_code_elimination` that deals with `constraint true` and `constraint false` inside of `const_folding`. And at that point it seems like we're going to end up in a pattern of copying and pasting the various optimization passes everywhere.

pintc/src/predicate/optimize/const_folding.rs
```diff
@@ -46,3 +47,218 @@ pub(crate) fn fold_consts(contract: &mut Contract) {
...
+/// false && <expr> is <false>
+/// x + 0 is x
+/// x - 0 is x
+/// x * 0 is 0
+/// x * 1 is x
+/// 0 / x is 0
+/// x / 1 is x
+/// x % 1 is x
+///
+/// Performs a depth-first search to find all folding opportunities.
```
lethalgem, line 66: Is this an appropriate way to document the identities we handle? Since we check for a specific conditions/values, I feel that it has to be commented somewhere. I'm not sure if it belongs in the function documentation or inline where we handle the identity in the function itself.

If we don't mark them, I can see it being a pain in the rear trying to add new cases in the future. Someone would have to figure out what is already there and then get to work adding more.

mohammadfawaz: Yeah this is fine :)

pintc/src/predicate/optimize.rs
mohammadfawaz: Yeah use the better ordering here... As we add more and more passes, you'll find that some passes need to be run multiple times in different places.

In practice, some orderings can be better than others. The rule we talked about emphasizes this statement:

> optimization passes need to operate independently and not depend on the order they're run in

This only means that each pass should be self contained and not dependent, **functionally**, on other passes. That is, you want the property that the passes _work_ (and do not crash etc.) regardless of which order they run in. That does not mean the produced code has to be exactly the same regardless of the order.

otrho:
I've taken a quick look at this now and it seems overly complicated for a few reasons.

I don't think `fold_identity()` needs to be recursive since you're visiting every expression anyway.  `fold_identity()` should just look at every binary expression and potentially replace it with one or other of its operands.  And only one of those operands (the one discarded) must be an immediate -- if both are immediates then they'll be constant folded elsewhere.  But also, as soon as you see `0 + a` you can just decide to replace the `+` with `a`, even if `a` _is_ an immediate.  This would imply perhaps that multiple passes would be needed to make sure all identity folds are found.

I think now that since the recursion (and testing?) is only on operands that are also operators then it's not comprehensive, and then not really needed.  I was looking at the code and wondering why it was matching on `(Ok(..), Err(..))` all the time, which looks a bit smelly to me.

Also, there's no need for a `HashMap` to keep the candidates, a `Vec<(ExprKey, ExprKey)>` would work and be cheaper.

So I think this can be simplified and still work.  Look at every binary op.  If it's an `&&` and the LHS is `true` then push the candidate.  Else if the RHS is `true` then push that.  Etc for other binary ops.  Replace candidates, rinse and repeat until the candidate list for a pass is empty.  I don't think you need to evaluate anything nor recurse, really?

Am I missing something here?

lethalgem:
> Also, there's no need for a `HashMap` to keep the candidates, a `Vec<(ExprKey, ExprKey)>` would work and be cheaper.

Good to know! I didn't realize a `vector of tuples` was cheaper than a `hashmap`.

> So I think this can be simplified and still work. Look at every binary op. If it's an `&&` and the LHS is `true` then push the candidate. Else if the RHS is `true` then push that. Etc for other binary ops. Replace candidates, rinse and repeat until the candidate list for a pass is empty. I don't think you need to evaluate anything nor recurse, really?
> 
> Am I missing something here?

Good catch. I didn't think of the fact that `fold_consts` is _replacing_ exprs with their evaluated immediates already.

I've also changed it so that we'll continuously run `fold_consts` and `fold_identities` until there are no more opportunities to fold. @mohammadfawaz enlightened me today with the knowledge that this is common practice during the optimization phase of a compiler.
```

File: pintc/src/predicate/optimize/const_folding.rs [hunk 5]
```diff
@@ -30,19 +66,203 @@ pub(crate) fn fold_consts(contract: &mut Contract) {
...
+                    Some(value)
+                } else {
+                    None
+                };
+
+                match op {
+                    BinaryOp::LogicalAnd => match (lhs_imm, rhs_imm) {
+                        (Some(Immediate::Bool(true)), _) => {
+                            expr_keys_to_replace.push((expr_key, *rhs));
+                        }
```
mohammadfawaz, line 131: Might be easier to consolidate the `match` statements?
```
                match (op, lhs_imm, rhs_imm) {
                    // Boolean identities
                    (BinaryOp::LogicalAnd, Some(Immediate::Bool(true)), _) => Some(*rhs),
                    (BinaryOp::LogicalAnd, _, Some(Immediate::Bool(true))) => Some(*lhs),
                    ...
                    // Arithmetic identities
                    (BinaryOp::Add, Some(Immediate::Int(0)), _) | (BinaryOp::Add, _, Some(Immediate::Int(0))) => Some(*rhs),
                    (BinaryOp::Add, Some(Immediate::Real(0.0)), _) | (BinaryOp::Add, _, Some(Immediate::Real(0.0))) => Some(*rhs),
```

mohammadfawaz: Something like this..

lethalgem: Updated

pintc/src/predicate/optimize/const_folding.rs
```diff
@@ -1,20 +1,56 @@
...
+        }
+
+        if fold_identities(contract) {
+            passes_that_folded += 1;
+        }
+
+        // There are no more folds to do.
+        if passes_that_folded == 0 {
+            break;
+        }
```
mohammadfawaz, line 30:
```suggestion
        if !fold_consts(contract) && !fold_identities(contract) {
            break;
        }
```

pintc/src/predicate/optimize/const_folding.rs
```diff
@@ -30,19 +66,131 @@ pub(crate) fn fold_consts(contract: &mut Contract) {
...
+
+                    (BinaryOp::Add, _, Some(Immediate::Int(0))) => Some(*lhs),
+
+                    (BinaryOp::Add, Some(Immediate::Real(0.0)), _) => Some(*rhs),
+
+                    (BinaryOp::Add, _, Some(Immediate::Real(0.0))) => Some(*lhs),
+
+                    (BinaryOp::Sub, _, Some(Immediate::Int(0))) => Some(*lhs),
+
+                    (BinaryOp::Sub, _, Some(Immediate::Real(0.0))) => Some(*lhs),
```
mohammadfawaz, line 154: We could technically replace `0 - x` or `0.0 - x` with `-x` right? We would have to insert a new `Expr::UnaryOp`

pintc/src/predicate/optimize/const_folding.rs
```diff
@@ -30,19 +66,131 @@ pub(crate) fn fold_consts(contract: &mut Contract) {
...
+
+                    (BinaryOp::Div, _, Some(Immediate::Int(1))) => Some(*lhs),
+
+                    (BinaryOp::Div, Some(Immediate::Real(0.0)), _) => Some(*lhs),
+
+                    (BinaryOp::Div, _, Some(Immediate::Real(1.0))) => Some(*lhs),
+
+                    (BinaryOp::Mod, _, Some(Immediate::Int(1))) => Some(*lhs),
+
+                    (BinaryOp::Mod, _, Some(Immediate::Real(1.0))) => Some(*lhs),
```
mohammadfawaz, line 174: Do we allow `Mod` for `real`? If so, that's a bug. The modulo operation should only be supported for integers.

pintc/src/predicate/optimize/const_folding.rs
```diff
@@ -30,19 +66,131 @@ pub(crate) fn fold_consts(contract: &mut Contract) {
...
+
```
mohammadfawaz, line 175: We could also add `1 % x = 1`

mohammadfawaz (commented):
Great! A few more comments :)

pintc/src/predicate/optimize/const_folding.rs
lethalgem: Yep, `Mod` is currently allowed for `real`. Will make it an issue and remove it from this PR

mohammadfawaz (approved):
Awesome :shipit: :)

PR #940, opened 2024-10-23
Predicate calls and remove pub vars

### LINKED ISSUE #934: Redesign `var` and `pub var`
* `var` and `pub var` become predicate args
* Can no longer add decision vars in macros. We should still be able to add `state` vars, but I noticed that right now we don't scope them. We probably should?

... To Be Continued

mohammadfawaz:
Closes #934

Plus lots of little things. 

Broad points:
1. Adds local and external predicate calls (parser, semantic checks, backend using `PredicateExists` etc.)
```rust
constraint LocalPredicate[[]](<arg1>, <arg2>, ..); // local
constraint ExternalInterface[[<addr>]]::ExternalPredicate[[<addr>]](<arg1>, <arg2>, ..); // external
```
3. Removes `pub var` everywhere including the parser, semantic checks, backend, abi and abi gen, etc.
4. Removes interface instances and predicate instances.
5. Change external storage accesses to directly using the syntax `[[addr]]` as in:
```rust
state x = ExternalInterface[[<addr>]]::x;
```
6. Predicate interfaces now lists their vars as arguments instead of a list `pub var`s between `{..}`


otrho (approved):
`+2,755 −4,277`  Nice.

PR #949, opened 2024-10-28
Update tests that use experimental features

lethalgem:
While working on #935 I noticed that a few of the tests when running `cargo test --features experimental-types` were failing

~~I guess we don't want to waste github actions on running the experimental tests, so these slipped under the radar in past PRs.~~

Actually we do run 
```
- name: Run Experimental Workspace Tests
        run: cargo nextest run --release --locked --features experimental-types -- experimental
```
in `.github/workflows/ci.yml` so I'm not sure how these made it through


lethalgem:
Nevermind. It appears that the remote does check and everything passes. On my local it does not pass and suggests changes.

PR #950, opened 2024-10-28
Add `salt` to `pint build` and `pintc`

### LINKED ISSUE #932: Add `pint build` support for specifying contract salt, e.g. `pint build --salt <hex>`
It should be possible to specify a salt when building a contract with `pint build`.

```console
pint build --salt 0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF
```

Attempting to provide a salt when building a library-only package should error.

### Workspace Support (#666) 

It should be possible to provide multiple salts in the case that `pint build` is building a workspace, so that in a single command, all packages can be built with their designated salts. This might look something like:

```console
pint build \
    --salt "foo:0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF" \
    --salt "bar:0101010101010101010101010101010101010101010101010101010101010101" \
    --salt "baz:1010101010101010101010101010101010101010101010101010101010101010"
```

where `foo`, `bar` and `baz` are names of contract packages in the workspace.

The `--salt` command-line argument should only be applied to the top-level contract members. That is, CLI-provided salts should not override the salts specified for contracts appearing in `[contract-dependencies]`. Instead, contract dependencies should have their salts specified in the `[contract-dependencies]` table entries (#931).

mohammadfawaz:
Closes #932 

* Adds `--salt` to `pint build`
* Adds `--salt` to `pintc`
* The salt value must exist and must be a string representing a 256-bit unsigned integer. Otherwise, an error is emitted. If less than 64 digits are provided, the salt value is left padded with zeros.
* If `--salt` is passed to `pint build` and the package being built is a library, then an error is emitted.

### DIFF
--- a/pint-abi-gen-tests/build.rs
+++ b/pint-abi-gen-tests/build.rs
@@ -12,7 +12,7 @@ fn build_test_pkg(manifest: ManifestFile) {
     let members = [(name, manifest)].into_iter().collect();
     let plan = pint_pkg::plan::from_members(&members).expect("failed to plan compilation");
     let built_pkgs = pint_pkg::build::build_plan(&plan)
-        .build_all(false /* skip_optimize */)
+        .build_all(Default::default(), false /* skip_optimize */)
         .expect("failed to build package");
 
     // Retrieve the built target package.
--- a/pint-cli/src/build.rs
+++ b/pint-cli/src/build.rs
@@ -2,7 +2,10 @@
 
 use anyhow::Context;
 use clap::{builder::styling::Style, Parser};
-use pint_pkg::{build::BuiltPkg, manifest::ManifestFile};
+use pint_pkg::{
+    build::BuiltPkg,
+    manifest::{ManifestFile, PackageKind},
+};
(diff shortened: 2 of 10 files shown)


otrho (dismissed):
Is it necessary for the salt to be a 64 character hex string always?  I feel  like it might be OK to allow any number on the command line, as long as it fits in 256bits.

mohammadfawaz:
> Is it necessary for the salt to be a 64 character hex string always? I feel like it might be OK to allow any number on the command line, as long as it fits in 256bits.

Good question. @mitchmindtree ?

mitchmindtree:
:thinking: I like the idea of not having to specify the whole thing and allow for more convenient small salts, but maybe there'd be ambiguity around whether the `0`s are padded to the start or end? I guess you'd think of it as a hex literal value so they'd always be implicitly padded to the start of the string.

I could go either way!

otrho (approved):

PR #953, opened 2024-10-30
Allow accessing `storage maps` in `tuples`

### LINKED ISSUE #808: Allow accessing storage maps in tuples
The following should work but doesn't:
```rust
storage {
    t: { ( int => int ), int },
}

predicate test {
    state t = storage::t.0[0];
}
```
Error:
```console
❯ pintc test.pnt
Error: expected `.`, or `;`, found `[`
   ╭─[test.pnt:6:27]
   │
 6 │     state t = storage::t.0[0];
   │                           ┬
   │                           ╰── expected `.`, or `;`
───╯
```

### LINKED ISSUE #952: `complex_map` example in Pint Book does not work
[This example](https://essential-contributions.github.io/pint/book/storage/dynamic.html#complex-maps) is either outdated or we have a bug in Pint now.

```
storage {
    complex_map: ( { int, int } => { bool, b256 } ),
}

results in 

Error: type not allowed in storage
   ╭─[test.pnt:4:18]
   │
 4 │     complex_map: ( { int, int } => { bool, b256 } ),
   │                  ─────────────────┬────────────────  
   │                                   ╰────────────────── found type ( {int, int} => {bool, b256} ) in storage
───╯
```

The error is generated here:
https://github.com/essential-contributions/pint/blob/c78a2bbae033ab1ec5bc12994d4d870f25ecb654/pintc/src/types.rs#L479-L505

Where the type check is performed on `Type::Map` and the `ty_from` is a `tuple`. This returns false for our check and the type is disallowed.

Should the Pint Book be updated or should the type checking be updated?

lethalgem:
Closes #808 
Closes #952

FTR, I hated everything about this issue. Thanks :)


pintc/tests/storage/bad_storage_types.pnt
```diff
@@ -2,38 +2,26 @@ union MyUnion = A | B;
...
-// @65..86: found type ( {int, int} => int ) in storage
+// @65..77: found type {int, int}[] in storage
 // type not allowed in storage
-// @95..110: found type ( int[5] => int ) in storage
-// type not allowed in storage
-// @119..134: found type ( int[] => int ) in storage
-// type not allowed in storage
-// @143..155: found type {int, int}[] in storage
-// type not allowed in storage
-// @164..178: found type ( int => int )[] in storage
```
lethalgem, line 38: Do we allow vectors in the map's `ty_from` field? I didn't see anything in the Pint Book that says otherwise.

The check below still requires that the vector be of type `bool`, `int`, or `b256`.

```rust
match self {
            Type::Map { ty_from, ty_to, .. } => {
                // We only support maps from these types for now
                (ty_from.is_bool() || ty_from.is_int() || ty_from.is_b256())
                (ty_from.is_bool()
                    || ty_from.is_int()
                    || ty_from.is_b256()
                    || ty_from.is_allowed_in_storage())
                    && ty_to.is_allowed_in_storage()
            }
            Type::Vector { ty, .. } => {
                // We only support vectors of these types for now
                ty.is_bool() || ty.is_int() || ty.is_b256()
            }
            Type::Array { ty, .. } => ty.is_allowed_in_storage(),
            Type::Tuple { fields, .. } => fields
                .iter()
                .fold(true, |acc, (_, field)| acc && field.is_allowed_in_storage()),
            Type::Alias { ty, .. } => ty.is_allowed_in_storage(),
            Type::Custom { .. } => {
                // Disallow custom types since they're ambiguous. Hopefully, by the time we need
                // this method, all custom types are gone.
                false
            }
            _ => true,
        }
...
```

mohammadfawaz: don't worry about vectors. They're not officially supported as they still need a lot of work.

lethalgem: Alright. So I should remove support and check directly for the approved tuple types?

pint-abi-gen-tests/test-pkgs/simple/src/contract.pnt
```diff
@@ -48,6 +49,7 @@ predicate Foo(
     let my_nested_map0_key1_2: int = mut storage::my_nested_map0[1][2];
     let my_nested_map1_key2_3: { int, { b256, int } } = mut storage::my_nested_map1[2][0x3333333333333333333333333333333333333333333333333333333333333333];
     let my_array: int[5] = mut storage::my_array;
+    let my_tuple_map: int = mut storage::my_tuple_map.0[0];
```
mohammadfawaz, line 52: This is good.. but I'd like to see more examples other than `.0[0]`

pintc/src/types.rs
```diff
@@ -483,7 +483,10 @@ impl Type {
         match self {
             Type::Map { ty_from, ty_to, .. } => {
                 // We only support maps from these types for now
-                (ty_from.is_bool() || ty_from.is_int() || ty_from.is_b256())
+                (ty_from.is_bool()
+                    || ty_from.is_int()
+                    || ty_from.is_b256()
+                    || ty_from.is_allowed_in_storage())
```
mohammadfawaz, line 489: So this is a bit incorrect since it would allow `ty_from` to be a map and that's never going to be allowed. I think I misled you in the issue. Instead, 

I think we can use the method `get_storage_only_ty` which, if returns `None`, then we know that the type can be used as a key. Then, we add tests for each case:
1. all primitive types.
2. tuples
3. unions
4. arrays
All as long as they don't _contain_ storage only types (which are basically maps and vectors only)

pintc/tests/storage/bad_storage_types.pnt
mohammadfawaz: Once you address my other comments about `get_storage_only_ty`, things should go back to normal.

mohammadfawaz:
When you're done, you can remove `<disabled>` from this test: https://github.com/essential-contributions/pint/blob/master/examples/ch_5_2_b.pnt

pintc/tests/storage/storage_map_keys.pnt
```diff
@@ -0,0 +1,46 @@
...
+    a: (int => b256),
+    b: (bool => int),
+    c: (b256 => bool),
+    d: ({ bool, int } => b256),
+    e: ({ int, b256 } => bool),
+    f: (MyUnion => int),
+    g: (int[5] => int),
+    h: ({ bool, b256 }[24][5] => int),
+    i: (MyAlias => bool),
+}
```
mohammadfawaz, line 14: Good.. now let's test these in `validation_tests` and `abi-gen-tests`.............
Sorry.........

lethalgem:
> When you're done, you can remove `<disabled>` from this example: https://github.com/essential-contributions/pint/blob/master/examples/ch_5_2_b.pnt
> 
> FYI, all examples are tested in CI. You can test them locally by running `cargo test` in `examples`. Examples that start with `// <disabled>` are not run of course.
> 
> This is where the code in #952 comes from. It is _included_ verbatim in the book using
> 
> https://github.com/essential-contributions/pint/blob/a939de9e856d922175686824f632fef447f0a2e8/docs/book/src/storage/dynamic.md?plain=1#L90

Good to know! It's been reenabled

pintc/tests/storage/storage_map_keys.pnt
lethalgem: just added the `validation_tests`. On to `abi-gen-tests`

otrho:
Sorry... merged some book changes and introduced conflicts.

mohammadfawaz (approved):
:shipit:

PR #958, opened 2024-10-31
Book updates after redesigning vars and pub vars

### LINKED ISSUE #938: Update the Book of Pint after the redesign of decision vars

mohammadfawaz:
Closes #938 

* Renamed the examples because keeping track of chapter numbers is a bit painful. Now, we don't need to change the names of the examples when the order of chapters change.
* Lots of rewording and updates related to parameters v.s. vars v.s. lets, predicate invocation v.s. pub vars, etc.
* Updates to macros. Some things need to be addressed and re-enabled (have issues open for that)


docs/book/src/examples/subcurrency.md
```diff
@@ -1,53 +1,53 @@
-## Subcurrency
+## subcurrency
```
otrho, line 1: The changes to this file are purely to make it all lowercase, AFAICT.  Is this deliberate?

otrho:
I pushed a little commit with some typo fixes, and a change of tone in one part -- I thought 'completely irrelevant' sounds a bit dismissive and used 'unimportant' instead.

docs/book/src/examples/subcurrency.md
mohammadfawaz: Oh damn it... must've changed everything to lowercase in vim by mistake.. let me fix.

otrho (approved):

PR #965, opened 2024-11-01
Add more `pint` printing flags

### LINKED ISSUE #956: Document the `--salt` flag in the Pint book

### LINKED ISSUE #960: Add `--print-parsed` to the `pint` CLI tool
Probably other flags too such as `--print-flat` and `--print-asm`.

`--print-parsed` in particular is important to inspect how macros are expanded.

mohammadfawaz:
Closes #956 
Closes #960 

Output of
```
 pint b --print-parsed --print-flat --print-optimized --print-asm
```
is
image

### DIFF
--- a/pint-cli/src/build.rs
+++ b/pint-cli/src/build.rs
@@ -26,10 +26,19 @@ pub(crate) struct Args {
     /// If "salt" is provided for a library package, an error is emitted.
     #[arg(long, value_parser = parse_hex)]
     salt: Option<[u8; 32]>,
-    /// Print the flattened pint program.
+    /// Print the parsed package.
+    #[arg(long = "print-parsed")]
+    print_parsed: bool,
+    /// Print the flattened package.
+    #[arg(long = "print-flat")]
+    print_flat: bool,
+    /// Print the optimized package.
     #[arg(long = "print-optimized")]
     print_optimized: bool,
-    /// Skip optimizing the pint program.
+    /// Print the assembly generated for the package.
+    #[arg(long = "print-asm")]
+    print_asm: bool,
+    /// Skip optimizing the package.
     #[arg(long = "skip-optimize", hide = true)]
     skip_optimize: bool,
     /// Don't print anything that wasn't explicitly requested.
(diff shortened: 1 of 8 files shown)


pint-cli/src/build.rs
```diff
@@ -102,6 +111,15 @@ pub(crate) fn cmd(args: Args) -> anyhow::Result<()> {
 
     // Build the given compilation plan.
     let mut builder = pint_pkg::build::build_plan(&plan);
+    let options = pint_pkg::build::BuildOptions {
+        salt: args.salt.unwrap_or_default(),
```
mitchmindtree, line 115: It looks like this may end up applying the salt to every package, but we only want to apply it to the "top-level" contract package that was targeted for building.

Rather than making this a single salt, one option could be to make this something liike `HashMap<Manifest, Salt>`, so that each package can check whether or not there was a custom salt specified for it during the build?

mitchmindtree (commented):
Nice one @mohammadfawaz , we might just wanna tweak how we pass down the salt, but otherwise I think the print stuff is probably right? in the case that we only want to apply any options to a single package now or in the future, maybe we could use the map idea mentioned in my other comment to pass through package-specific options more generally?

pint-cli/src/build.rs
mohammadfawaz: Alright, I addressed this in https://github.com/essential-contributions/pint/pull/965/commits/63648b9dabf1c65bf1f97c4060ef578b4bf08fa7.. Please let me know if this makes sense!

pint-pkg/src/build.rs
```diff
@@ -41,7 +41,7 @@ pub struct BuildOptions {
     /// contract unique.
     ///
     /// If "salt" is provided for a library package, an error is emitted.
-    pub salt: [u8; 32],
+    pub salts: HashMap<manifest::ManifestFile, [u8; 32]>,
```
mitchmindtree, line 44: It might be enough to make this `manifest::Manifest` instead of using `ManifestFile` (`ManifestFile` derefs to `Manifest` btw).

I'm still not 100% using `Manifest` as a key like this is the best approach, but we need something that is reliably unique, so I guess it will do for now? In the build plan we use `pint_pkg::plan::PinnedId` for this, but it's quite specific to the planning step.

mitchmindtree (approved):
:shipit:

pint-pkg/src/build.rs
mohammadfawaz: I'll open an issue to track this :) Thanks!

PR #969, opened 2024-11-04
Fix macros and `let`.

### LINKED ISSUE #937: Make `state` declarations hygenic in macros
Now that `state` can be initialized to anything, macros should be able to make them hygenic, just like we currently do with `var`s. After https://github.com/essential-contributions/pint/pull/933, we won't be able to do that for vars, so only `state` should support that behavior.

### LINKED ISSUE #954: Array splicing for `let` vars
Similar to https://github.com/essential-contributions/pint/issues/937, we need to allow array splicing for `let` vars.

otrho:
Closes #937.  Closes #954.


mohammadfawaz (approved):
Awesome, thanks!

PR #989, opened 2024-11-14
Update interface and predicate instance syntax

mohammadfawaz:
Change from `[[...]]` to `@[...]`


otrho (approved):
Yeah, this is obviously better for both the parser and for readability.

I'd be tempted to make `@[` the opening token rather than `@` `[` to further reduce the possibility of ambiguity (at the moment you could put a newline after the `@` and it'd be fine, which maybe it shouldn't be) and maybe more informative syntax error messages, but this obviously works and is straight forward. :+1:

PR #1000, opened 2024-11-16
Refactor internal errors

### LINKED ISSUE #829: Replace all the `assert!()`s in `asm_gen.rs` with `InternalError`s.
And potentially:
- replace the message `&str` with a `String` which can be formatted to contain error context.
- add a method or macro to streamline creating `InternalError` results.

mohammadfawaz:
Closes #829 (the part about asm gen `asserts` is no longer relevant since there are no more asserts there but I removed a few other things that could panic)
* Move `Internal` to `Error` instead of `CompileError`. 
* Add `emit_internal_err` to `Handler`


otrho (approved):
There's a lot of internal errors. :thinking:

PR #1005, opened 2024-11-19
Disallow multiple identical predicates in the same contract

### LINKED ISSUE #848: Disallow multiple identical predicates in the same contract
That causes a clash in their addresses.

lethalgem:
Closes #848

### DIFF
--- a/pintc/src/asm_gen.rs
+++ b/pintc/src/asm_gen.rs
@@ -1,13 +1,14 @@
 use crate::{
-    error::{ErrorEmitted, Handler},
+    error::{CompileError, Error, ErrorEmitted, Handler},
     expr::{Expr, ExternalIntrinsic, Immediate, IntrinsicKind},
     predicate::{ConstraintDecl, Contract, Predicate},
     span::empty_span,
 };
 use asm_builder::AsmBuilder;
 use essential_types::{predicate::Predicate as CompiledPredicate, ContentAddress};
+use fxhash::FxHashMap;
+use fxhash::FxHashSet;
 use petgraph::{graph::NodeIndex, Graph};
-use std::collections::HashMap;
 
 mod asm_builder;
 mod display;
@@ -33,11 +34,11 @@ pub fn compile_contract(
 
     // This map keeps track of what node indices (in the dependency graph) are assigned to what
     // predicates (identified by names)
-    let mut names_to_indices = HashMap::<String, NodeIndex>::new();
+    let mut names_to_indices = FxHashMap::<String, NodeIndex>::default();
 
     // This is a map between node indices in the dependency graph and references to the
     // corresponding predicates.
-    let mut indices_to_predicates = HashMap::<NodeIndex, &Predicate>::new();
+    let mut indices_to_predicates = FxHashMap::<NodeIndex, &Predicate>::default();
 
     for (_, pred) in contract.preds.iter() {
         let new_node = dep_graph.add_node(pred.name.clone());
@@ -85,8 +86,8 @@ pub fn compile_contract(
     // This map keeps track of the compiled predicates and their addresses. We will use this later
     // when producing the final compiled contract. It is also useful when compiling predicates that
     // require the addresses of other predicates.
-    let mut compiled_predicates: HashMap<String, (CompiledPredicate, ContentAddress)> =
-        HashMap::new();
+    let mut compiled_predicates: FxHashMap<String, (CompiledPredicate, ContentAddress)> =
+        FxHashMap::default();
 
     // Now compile all predicates in topological order
     for idx in &sorted_nodes {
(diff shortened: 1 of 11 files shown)


pintc/src/asm_gen.rs
```diff
@@ -103,6 +104,28 @@ pub fn compile_contract(
...
+                    original_name: original_name.to_string(),
+                    duplicate_name: string.to_string(),
+                    span: empty_span(),
+                },
+            });
+        } else {
+            original_predicate_names.insert(content_address, string);
+        }
+    }
+
```
lethalgem, line 128: @mohammadfawaz @otrho Do ya'll know if we have easy access to the predicate's name span? 

Right now the error produced is

```
predicate Foo() {
}

predicate Bar() {
}

predicate Beast(x: int) {
    constraint x > 2;
}

predicate Animal(x: int) {
    constraint x > 2;
}

results in 

Error: Identical predicates found in the same contract
   ╭─[:1:1]
   │
 1 │ <none>
   │ │ 
   │ ╰─ predicates `::Beast` and `::Animal` have identical constraints
   │ 
   │ Note: contract cannot compile unless each predicate has a unique overall set of constraints, even if individual constraints within the set are not unique
───╯
Error: Identical predicates found in the same contract
   ╭─[:1:1]
   │
 1 │ <none>
   │ │ 
   │ ╰─ predicates `::Bar` and `::Foo` have identical constraints
   │ 
   │ Note: contract cannot compile unless each predicate has a unique overall set of constraints, even if individual constraints within the set are not unique
───╯
```

Looking at `predicate` and `compiled_predicate` we don't appear to keep track.

I'll also take a look to see if we have any errors that don't require a span and follow suit temporarily.

lethalgem: We keep track of the `span` for any `param` for the `predicate`. That's the closest I've seen to the location of the `name` of the `predicate`.

otrho: Or if not we could just stuff the predicate name span into the predicate from the parser.

lethalgem: I had to make `symbols` public

https://github.com/essential-contributions/pint/blob/6aa8453946d8acba661d25433fd0889c932bb6a3/pintc/src/predicate.rs#L988-L991

If that's not alright, then I can do as @otrho suggests and add the info to our `predicate` struct

pintc/src/asm_gen.rs
```diff
@@ -103,6 +104,38 @@ pub fn compile_contract(
...
+            let original_span = contract
+                .symbols
+                .symbols
+                .get(*original_name)
+                .expect("original predicate name missing in contract symbols table");
+            let span = contract
+                .symbols
+                .symbols
+                .get(string)
+                .expect("predicate name missing in contract symbols table");
```
mohammadfawaz, line 124: This really is a pain.. Starting to think @otrho's suggestion is better. So yeah you'd have to change the parser to collect an `Ident` for `Predicate` instead of a `String`. Then, the span is readily available through `.span`.

lethalgem: https://github.com/essential-contributions/pint/blob/04094621349fcf42e31bb6e8bee053f66f7f07c4/pintc/src/predicate.rs#L956-L961

Is this what you had in mind? 

it results in a lot of `pred.name.name` everywhere that used to use a `string`. Maybe not an issue?

Or did you want me to stuff span in to the struct itself ala 
```
pub struct Predicate {
    pub name: String,
    pub span: Span,
    ...
```

pintc/src/error/compile_error.rs
```diff
@@ -372,6 +372,13 @@ pub enum CompileError {
     },
     #[error("invalid position for accessing storage")]
     InvalidStorageAccess { span: Span },
+    #[error("Identical predicates found in the same contract")]
```
mohammadfawaz, line 375: lower case `I`in `Identical` to follow the rest of the errors.

pintc/src/error/compile_error.rs
```diff
@@ -1229,6 +1236,23 @@ impl ReportableError for CompileError {
...
+                span,
+                ..
+            } => vec![
+                ErrorLabel {
+                    message: "original declaration here".to_string(),
+                    span: original_span.clone(),
+                    color: Color::Red,
+                },
+                ErrorLabel {
+                    message: "identical constraints here".to_string(),
```
mohammadfawaz, line 1250:
```suggestion
                    message: "identical predicate here".to_string(),
```

pintc/src/error/compile_error.rs
```diff
@@ -1355,6 +1379,10 @@ impl ReportableError for CompileError {
                     .to_string(),
             ),
 
+            IdenticalPredicates { .. } => {
+                Some("predicates may share some, but not all constraints".to_string())
```
mohammadfawaz, line 1383:
```suggestion
                Some("two predicates in a contract cannot have the exact same bytecode".to_string())
```

pintc/src/error/compile_error.rs
lethalgem: I worry that this may not be clear to the user. 

Ex.
```
predicate Foo() {
}
predicate Bar() {
    let y = 2;
}
```
will still pop the error.

mohammadfawaz: Yeah good point. The trouble is that, while constraints may seem identical, some `let`s may be different and that affects the addresses. So we need a more generalized way of indicating what's going on.

maybe:
"predicate with identical bytecode here"

and then in the note:

"two predicates in a contract cannot have the exact same (optimized) bytecode"

mohammadfawaz (approved):

PR #1009, opened 2024-11-26
Add type checks to `const` initialisers.

### LINKED ISSUE #913: type check before evaluating `consts`
`const f = true in 3..2` will make it to the evaluation stage even though it _should_ result in 

```
Error: value type and array element type in range differ
   ╭─[test.pnt:7:22]
   │
 7 │     const f = true in [1, 2];
   │                        ┬
   │                        ╰── array element type mismatch; expecting `bool` type, found `int` type
───╯
```

> ...
> A user can still write `const f = true in 3..2;` without an error occurring before this stage.
> 
> FWIW, if it is _not_ a `const` then we type check the range. Ex.
> ```
> predicate Test {
>     var l = true in 1..2;
> }
> 
> Error: value type and range type differ
>     ╭─[test.pnt:18:21]
>     │
>  18 │     var l = true in 1..2;
>     │                     ┬
>     │                     ╰── range type mismatch; expecting `bool` type, found `int` type
> ────╯
> ```
>
> _Originally posted by @lethalgem in https://github.com/essential-contributions/pint/pull/911#discussion_r1790795496_

otrho:
This was more painful than expected.  Constant initialiser evaluation is done before type checking so that we may use consts for array ranges (and other places?  I think arrays is it...) This change introduces a preliminary type check for constant initialisers before they are evaluated, mostly so we can catch bad types, as per #913.

But sometimes there will be type errors just because the consts are declared out of order and have dependencies.  That's why we iterate on evaluating them until they all resolve.  So those type errors are bogus and should be thrown away.  The approach I've taken in this PR was a compromise between a few different approaches I attempted and seems to balance out ensuring we report bad types but also not to get too verbose.

Unfortunately it introduces error duplication which are found during the type check but then also during the evaluation.  But I have another PR to address that.

Closes #913.

### DIFF
--- a/pintc/tests/evaluator/in_exprs.pnt
+++ b/pintc/tests/evaluator/in_exprs.pnt
@@ -1,6 +1,6 @@
 const a = 1 in [1, 2];
 const b = 1 in [3, 2];
-const c = true in [3,2];
+//const c = true in [3,2];
 const d = true in [true, false];
 const e = 1 in 1..2;
 const f = 1 in 3..2;
@@ -32,13 +32,12 @@ predicate Test(
 }
 
 // parsed <<<
-// const ::f = 1 in 3..2;
 // const ::b = 1 in [3, 2];
 // const ::e = 1 in 1..2;
 // const ::d = true in [true, false];
 // const ::a = 1 in [1, 2];
-// const ::c = true in [3, 2];
-// 
+// const ::f = 1 in 3..2;
+//
 // predicate ::Test(
 //     ::g: bool,
 //     ::h: bool,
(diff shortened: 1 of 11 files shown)


otrho:
I am going insane.

pintc/tests/evaluator/in_exprs.pnt
```diff
@@ -1,6 +1,6 @@
 const a = 1 in [1, 2];
 const b = 1 in [3, 2];
-const c = true in [3,2];
+//const c = true in [3,2];
```
mohammadfawaz, line 3: Is this intentional?

mohammadfawaz (approved):

PR #1010, opened 2024-11-26
Change handler to emit only a single error per span.

otrho:
I think reporting more than one error per span, even if they're strictly different errors, but especially if they're the same error, is noisy and generally a poor user experience.

So here is a patch which filters errors and warnings uniquely based on their span when extracting them from their `Handler`.  Looking at the changes in tests it looks worthwhile, especially after my last PR which ends up calling the type checker multiple times on consts and can duplicate errors.

But even for multiple errors per span which differ, usually the first one is enough and the others are superflous.  But YMMV and this PR doesn't need to be merged necessarily.

### DIFF
--- a/pintc/src/error/handler.rs
+++ b/pintc/src/error/handler.rs
@@ -110,8 +110,26 @@ impl Handler {
     }
 
     pub fn consume(self) -> (Vec<Error>, Vec<Warning>) {
-        let inner = self.inner.into_inner();
-        (inner.errors, inner.warnings)
+        use super::Spanned;
+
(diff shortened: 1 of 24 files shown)


lethalgem (dismissed):
I'm all for it. Looking at the changes in the test results, it declutters a lot.

And I don't think I've ever experienced an LSP reporting multiple errors for a particular span either, so this must be somewhat common practice.

mohammadfawaz (dismissed):

mohammadfawaz (approved):

PR #1011, opened 2024-11-26
`Fold` known `let` variables

### LINKED ISSUE #946: Optimizations around `let` variables
In the future, we want do more involved things such as removing unnecessary lets (const folding, DCE, etc.). That is, those variables don't necessarily _need_ a state program. So, someone could write:
```rust
let x = 5;
let y = storage::y + x;
```
the first `let` shouldn't need a state program while the second should.

lethalgem:
_DRAFT_ Tests are currently failing due to mismatched hardcoded contract and predicate addresses.

------
Closes #946


tests/validation_tests/storage_tuple.solution.json
```diff
@@ -2,8 +2,8 @@
   "data": [
     {
       "predicate_to_solve": {
-        "contract": "B209AF469914AA6DE8ECA843A31351712FFA491AAE71CE2CD9F935597EBB70E9",
-        "predicate": "3D7D302EE1E82A25B262B6F6F728E0A66E75A61D0C6A78AB5AD913796AE049A5"
+        "contract": "6F7917A688D243501B23C847DF70A21C32AF5868E3CEBD96A3D9056021D89CA1",
+        "predicate": "4B1CF45F80623E58BD0C7B90989B42F9BBB0C054D33C8181170C22E078DE3C2C"
```
lethalgem, line 6: Any idea why I'm not able to hash the proper addresses for the contract and predicates in `validation_e2e` tests?

I've used the local `pint build` to get the hash of the bytecode for `storage_tuple.pnt`, and I'm still getting the error `Error: solution for test validation_tests/storage_tuple.pnt does not solve any predicates in the test!`

I jumped back to the `master` branch and I'm unable to reproduce the original hashes too. Even weirder, I get 
```
"contract": "6F7917A688D243501B23C847DF70A21C32AF5868E3CEBD96A3D9056021D89CA1",
"predicate": "4B1CF45F80623E58BD0C7B90989B42F9BBB0C054D33C8181170C22E078DE3C2C"
```
for both this branch and `master`.

mohammadfawaz: I pushed a fix for `storage_tuple` that updates the addresses. You need to do the same for the other tests one by one. Make sure you're compiling `pint` and using the build. When you run the tests, the binary `pint` is not actually updated. You need to run `cargo build` separately.

lethalgem: >... You need to run cargo build separately.

This was the issue. 🤦 Thank you sir!

pintc/src/predicate/optimize/const_folding.rs
```diff
@@ -70,6 +82,7 @@ pub(crate) fn fold_consts(contract: &mut Contract) -> bool {
             did_fold_const = true;
         }
         expr_keys_to_replace = FxHashMap::default();
+        scope_values = FxHashMap::default();
```
mohammadfawaz, line 85: So might as well declare `scope_values` inside the `for` loop to avoid paying the cost of clearing it.

pintc/src/predicate/optimize/const_folding.rs
```diff
@@ -33,13 +33,25 @@ pub(crate) fn const_folding(handler: &Handler, contract: &mut Contract) {
...
-
     for pred_key in contract.preds.keys().collect::<Vec<_>>() {
+        // Create a map of variables that were declared as immediates for folding later
+        if let Some(pred) = contract.preds.get(pred_key) {
+            for var in pred.variables() {
+                let expr = var.1.expr.get(contract);
+                if let Expr::Immediate { value, .. } = expr {
+                    scope_values.insert(var.1.name.clone(), value.clone());
+                }
+            }
```
mohammadfawaz, line 50: Something like this I find a bit more readable. I try to avoid `.0`, `.1`, etc. if possible because it's not immediately clear what they refer to. 

Also, I tend to avoid temporary variables if they're only used once and if they're not initialized with a "large" RHS.
```suggestion
            for (_, var) in pred.variables() {
                if let Expr::Immediate { value, .. } = var.expr.get(contract) {
                    scope_values.insert(var.1.name.clone(), value.clone());
                }
            }
```

mohammadfawaz (commented):
Great! Nothing major, just some small suggestions.

PR #1018, opened 2024-11-28
Prevent appending whitespace to test comments when using UPDATE_EXPECT.

otrho:
This is a personal thing I imagine -- I have both a highlighter for wayward whitespace in my Neovim setup, and an autoformatter which removes it.  It's probably just anal retention but I don't like whitespace at the end of lines.

The super useful `UPDATE_EXPECT` feature we have for tests was adding them though, for blank lines, emitting a `// ` and getting highlighted by my Neovim and also automatically removed and just causing weird diffs in PRs...

So here's a tiny patch to stop emitting the whitespace.  But it's also a massive patch cos it's 'fixing' a whole lot of our current test files.

Feel free to discard this PR if it's silly.


mohammadfawaz (approved):
Yeah this makes sense.. I don't know why it was adding the whitespace initially 🤷

PR #1023, opened 2024-12-02
Fix for Rust v1.83.0.

otrho:
1.83.0 introduced a few new warnings about missing lifetimes and clippy became more strict about unused named lifetimes and gaps after attributes.


otrho:
CI is still on v1.81.0.

mohammadfawaz (approved):

PR #1026, opened 2024-12-04
Print content addresses when compiling with `pintc`

### LINKED ISSUE #1012: Return content addresses with `pintc` instead of just `pint cli`
It's getting to be pretty annoying spinning up a new contract with `pint cli` just to get the address for a `contract` or `predicate`.

lethalgem:
Close #1012


pintc/tests/cli.rs
```diff
@@ -142,8 +156,17 @@ fn explicit_output() {
...
+    let expected_output = format!("    contract {input_file_name}       EB87FCE275A9AB10996D212F39221A56B90E01C37FA9D16EE04A3FE8E17DEED9
+         └── {input_file_name}::test BA6595C5C75346E6C82BED0CE770D0758ADD1712163FCE45E38E5E8EAC6AA153\n");
+    write!(expected_output_file.as_file_mut(), "{}", expected_output).unwrap();
+
     check(&output.stderr, expect_test::expect![""]);
-    check(&output.stdout, expect_test::expect![""]);
+    check_file(
+        &output.stdout,
+        expect_test::expect_file![expected_output_file.path()],
+    );
```
lethalgem, line 169: Super ugly. It was the only way I could get the tests to work properly while still allowing `trunk fmt` to run. I'd like it to match the output directly like 
```
r#"
    contract {input_file_name}  EB87FCE275A9AB10996D212F39221A56B90E01C37FA9D16EE04A3FE8E17DEED9
    └── {input_file_name}::test BA6595C5C75346E6C82BED0CE770D0758ADD1712163FCE45E38E5E8EAC6AA153
"#
```
Which I can't do if I want to write to a file.

I also had to write to a file because the `expect![` macro was removing all whitespace at the beginning of the `string literal`, even when it was a `raw` string. It also meant I couldn't use the dynamic tempfile filename. Hardcoding it wasn't an issue, it just meant that it was more to deal with when creating a test.

Open to ideas here. Spent a few hours on it and have decided that life wasn't meant to be perfect.

pintc/src/main.rs
```diff
@@ -103,6 +103,49 @@ fn main() -> anyhow::Result<()> {
...
+                let mut name_w = 0;
+                for name in &compiled_contract.names {
+                    let w = name.chars().count();
+                    name_w = std::cmp::max(name_w, w);
+                }
+                pinned_name.chars().count() + name_w
+            };
+            let bold = Style::new().bold();
+            println!(
+                "    {}contract{} {:<name_col_w$} {}",
```
lethalgem, line 132: This will always be `contract`, correct?

pintc/src/main.rs
```diff
@@ -103,6 +103,49 @@ fn main() -> anyhow::Result<()> {
...
+
+            let mut iter = compiled_predicate_contract_addresses
+                .iter()
+                .zip(&compiled_contract.names)
+                .peekable();
+            while let Some((ca, name)) = iter.next() {
+                let name = format!("{}{}", pinned_name, name);
+                let pipe = iter.peek().map(|_| "├──").unwrap_or("└──");
+                println!("         {pipe} {:<name_col_w$} {}", name, ca);
+            }
```
lethalgem, line 147: I added quite a bit to `main.rs` to print the addresses properly. I do think it's pretty straightforward/clean, so maybe it's not a big deal? 

Let me know if you'd like me to move it elsewhere. I'd probably opt to create a struct with the sole purpose of using the `display` trait to keep everything organized.

pintc/tests/cli.rs
```diff
@@ -121,8 +126,17 @@ fn default_output() {
     assert!(input_file.path().with_extension("json").exists());
     let _ = fs::remove_file(input_file.path().with_extension("json"));
 
+    let input_file_name = input_file.path().file_name().unwrap().to_str().unwrap();
+    let mut expected_output_file = tempfile::NamedTempFile::new().unwrap();
+    let expected_output = format!("    contract {input_file_name}       EB87FCE275A9AB10996D212F39221A56B90E01C37FA9D16EE04A3FE8E17DEED9
+         └── {input_file_name}::test BA6595C5C75346E6C82BED0CE770D0758ADD1712163FCE45E38E5E8EAC6AA153\n");
```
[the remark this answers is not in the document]
lethalgem: No :(

I was unable to find a way to get that to work. 

When it was a string literal in `expect![` it could be updated, but would fail the next time the test was run. This is because the whitespace in front of `        contract` was being ignored even with a `raw string`.

If you know a way to preserve the whitespace in the string without the macro affecting it, then all problems would be solved :)

pintc/src/main.rs
otrho: IMO, `main()` is already a bit big.  It's probably possible to split it up into different functions, though doing that for the sake of it is subjective.  I.e., is it actually clearer when split up?  Is it clearer with every stage after each other?

Using `Display` isn't really necessary though, IMO.  Especially if it's pretty much only used once.

pintc/tests/cli.rs
otrho: Does it absolutely, truly matter if the whitespace is a little off in the test?  What if, as a hack for convenience, we use the `expect!` as usual, but throw in an ancillary `assert_eq!(output[0..4], "    ")` or whatever, which forever ensures the whitespace is there, pinky promise.

pintc/tests/cli.rs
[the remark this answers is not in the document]
mohammadfawaz: Ok so here's my suggestion:
```rust
    let stdout = output.stdout.replace(
        input_file.path().file_stem().unwrap().to_str().unwrap(),
        "<filename>",
    );
    let stdout = stdout.trim();

    check(
        &stdout,
        expect_test::expect![[r#"
        contract <filename>       EB87FCE275A9AB10996D212F39221A56B90E01C37FA9D16EE04A3FE8E17DEED9
                 └── <filename>::test BA6595C5C75346E6C82BED0CE770D0758ADD1712163FCE45E38E5E8EAC6AA153"#]],
    );
```
This seem to work well. Updates with `expect-test` and does not care about trailing spaces.

pintc/tests/cli.rs
```diff
@@ -10,6 +10,11 @@ fn check(actual: &str, expect: expect_test::Expect) {
     expect.assert_eq(actual);
 }
 
+#[cfg(test)]
+fn check_file(actual: &str, expect: expect_test::ExpectFile) {
+    expect.assert_eq(actual);
+}
```
mohammadfawaz, line 16: With my suggestion below, this shouldn't be needed anymore.

pintc/src/main.rs
mohammadfawaz: Yeah pretty much.. for now.

pintc/src/main.rs
mohammadfawaz: Yeah I guess it's fine for now.. you can open an issue if you want to refactor main into smaller pieces.. but it's low priority.

lethalgem: I'll hold off on opening an issue. Seems like it'd just be bikeshedding at best.

mohammadfawaz (approved):
:shipit:

PR #1029, opened 2024-12-06
Change the strictness around shadowing for match bindings.

### LINKED ISSUE #879: Unions: Minor fixes and/or updates.

otrho:
We don't allow shadowing anywhere else, so it's consistent to disallow them for match bindings too.

Closes #879.


mohammadfawaz (dismissed):

mohammadfawaz:
Oh CI failures...

mohammadfawaz (dismissed):

otrho:
Dammit, you gotta be quick.  Else your PR is sabotaged. :wink:

mohammadfawaz (approved):

PR #1035, opened 2024-12-09
Update `pintc` to 0.13.0.

otrho:


mitchmindtree (approved):
Lgtm

nfurfaro (dismissed):
Looks good!

mitchmindtree (changes requested):
Actually I think we also need to bump pint-cli - we should bump everything that has the essential-* crates even as transitive dependencies, just to be safe in case they're publicly exposed in some way

mitchmindtree (approved):

PR #1036, opened 2024-12-09
Wrong type for next state tuple field access

### LINKED ISSUE #1014: Wrong type for next state tuple field access
Example:
```rust
storage {
    t: {int, int},
}

predicate foo() {
    let t = storage::t;
    constraint t.0' == 0;
}
```
This compiles to:
```rust
storage {
    t: {int, int},
}

predicate ::foo(
) {
    let ::t: {int, int} = __storage_get({0, 0});
    constraint (::t'.0 == 0);
}
```
but then we think the type of `::t'` is `int`. This does not happen if we write `constraint t'.0 == 0` in the original code.

lethalgem:
Closes #1014

### DIFF
--- a/pintc/src/asm_gen/asm_builder.rs
+++ b/pintc/src/asm_gen/asm_builder.rs
@@ -1104,8 +1104,9 @@ impl<'a> AsmBuilder<'a> {
 
         // Grab the fields of the tuple
         let Type::Tuple { ref fields, .. } = tuple_expr.get_ty(contract) else {
-            return Err(handler
-                .emit_internal_err("type must exist and be a tuple type", empty_span()));
+            return Err(
+                handler.emit_internal_err("type must exist and be a tuple type", empty_span())
+            );
         };
 
         // The field index is based on the type definition
--- a/pintc/src/asm_gen/asm_builder.rs
+++ b/pintc/src/asm_gen/asm_builder.rs
@@ -1105,7 +1105,7 @@ impl<'a> AsmBuilder<'a> {
         // Grab the fields of the tuple
         let Type::Tuple { ref fields, .. } = tuple_expr.get_ty(contract) else {
             return Err(handler
-                .emit_internal_err("type must exist and be a tuple type - ian", empty_span()));
+                .emit_internal_err("type must exist and be a tuple type", empty_span()));
         };
 
         // The field index is based on the type definition
--- a/pintc/src/asm_gen/asm_builder.rs
+++ b/pintc/src/asm_gen/asm_builder.rs
@@ -1107,7 +1107,6 @@ impl<'a> AsmBuilder<'a> {
         }
 
         // Grab the fields of the tuple
-        let Type::Tuple { ref fields, .. } = tuple.get_ty(contract) else {
         println!("after compiling pointer");
         println!("key for tuple expr is: {:#?}", tuple);
         println!("tuple expr is: {}", contract.with_ctrct(tuple));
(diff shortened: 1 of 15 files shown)


pintc/src/asm_gen/asm_builder.rs
```diff
@@ -1090,11 +1090,22 @@ impl<'a> AsmBuilder<'a> {
...
+            op: UnaryOp::NextState,
+            expr,
+            ..
+        } = tuple.get(contract)
+        {
+            expr
+        } else {
+            tuple
+        };
+
```
lethalgem, line 1104: @otrho While I add e2e tests, I wanted to double-check the fix logic with you.


https://github.com/essential-contributions/pint/blob/8ea90ce77b392397e0d6c7b298e9c13500f44905/pintc/src/predicate/transform/lower.rs#L1074-L1102
Here the `UnaryOp::NextState` is stuffed into the `Expr::TupleFieldAccess` with `*old_accessed_key = op_key;`. I _believe_ that is correct, as we want to know that the field access is a prime op. If that's the case, then we want to make sure to dig into the field access and see if the `Expr::Tuple` is inside of the prime op before attempting to access the fields of the tuple.

Does that seem correct? Or should the prime op not be inside the `Expr::TupleFieldAccess` and be preserved elsewhere?

otrho: Yeah, this is a bit unintuitive.  I think the prime op should be as close to the base storage value as possible, so it would go inside the field access.  This way, when we're fetching storage into state, it's known without any fuss whether we want pre- or post-.

So the type of the field access is the type of the field, and if the field is a prime op, it should be transparent.  I haven't looked deeper into why we're getting those internal errors though, I'll leave that to you. :slightly_smiling_face:

lethalgem: > Yeah, this is a bit unintuitive. I think the prime op should be as close to the base storage value as possible, so it would go inside the field access. This way, when we're fetching storage into state, it's known without any fuss whether we want pre- or post-.

So the current setup is correct then 👍 

> So the type of the field access is the type of the field, and if the field is a prime op, it should be transparent. I haven't looked deeper into why we're getting those internal errors though, I'll leave that to you. 🙂

We were getting the error because the prime op was the field access, as intended. When in the `asm_builder`, we just didn't check if it was a prime op first. So instead of digging into the prime op to get the tuple that the prime op references, we panicked. The fix above should remedy that.

Here's where we got the error before I made the changes
```
let Type::Tuple { ref fields, .. } = tuple.get_ty(contract) else {
    return Err(
    handler.emit_internal_err("type must exist and be a tuple type", empty_span())
    );
```

pintc/src/asm_gen/asm_builder.rs
```diff
@@ -1090,11 +1090,22 @@ impl<'a> AsmBuilder<'a> {
...
+        let tuple_expr = if let Expr::UnaryOp {
+            op: UnaryOp::NextState,
+            expr,
+            ..
+        } = tuple.get(contract)
+        {
+            expr
+        } else {
+            tuple
+        };
```
otrho, line 1103: It feels like this is too specific and we should have a more general solution for getting the tuple expression, but I guess as this stage any other 'wrapper' has been lowered so it's OK.

OTOH, perhaps we should be pre-resolving state accesses as pre- or post- and then removing the next-state ops altogether.

So this is fine for now, but I think it might not be a long term solution.

pintc/src/asm_gen/asm_builder.rs
```diff
@@ -1090,11 +1090,22 @@ impl<'a> AsmBuilder<'a> {
...
+
         // Grab the fields of the tuple
-        let Type::Tuple { ref fields, .. } = tuple.get_ty(contract) else {
-            return Err(
-                handler.emit_internal_err("type must exist and be a tuple type", empty_span())
-            );
+        let Type::Tuple { ref fields, .. } = tuple_expr.get_ty(contract) else {
+            return Err(handler
+                .emit_internal_err("type must exist and be a tuple type - ian", empty_span()));
```
otrho, line 1108:
```suggestion
                .emit_internal_err("type must exist and be a tuple type", empty_span()));
```

otrho (approved):

PR #1038, opened 2024-12-18
Minor refactor of asm gen

mohammadfawaz:
1. Minor refactor of how data is collected and returned for non-leaf nodes. This reduces the size of the asm quite a bit as can be seen in the tests.
2. Remove `compile_expr_pointer_deref` as it's no longer necessary with this change.
3. Use short names for ops.


mohammadfawaz:
@otrho if this highly conflicts with your current work, we don't need to merge it. It's nice to have but not critical.

pintc/src/asm_gen/asm_builder.rs
```diff
@@ -820,28 +734,18 @@ impl<'a> AsmBuilder<'a> {
...
-                        asm.push(Op::Crypto(Crypto::RecoverSecp256k1))
-                    }
+                    ExternalIntrinsic::RecoverSECP256k1 => asm.push(RSECP),
 
-                    ExternalIntrinsic::Sha256 => {
-                        asm.push(Op::Stack(Stack::Push(8))); // placeholder for index
-                        asm.push(Alu::Mul.into());
-                        asm.push(Op::Crypto(Crypto::Sha256))
-                    }
+                    ExternalIntrinsic::Sha256 => asm.extend([PUSH(3), SHL, SHA2]),
```
mohammadfawaz, line 744: `SHL` is generally going to be more efficient than `MUL`

otrho (approved):

PR #1039, opened 2024-12-20
Allow storage accesses everywhere

### LINKED ISSUE #979: Allow storage accesses everywhere
Description is WIP
1. Allow storage accesses everywhere
2. Allow the prime (') operator on storage accesses (as in `constraint storage::counter' == storage::counter + 1`
3. Probably disallow `'` everywhere else for simplicity (?)

This means that `let` declarations are no longer required for storage accesses, but only for code organization etc.

Work on this is pending support for compute graph in asm gen because it also requires changes to the `KeyRange` opcode etc.

mohammadfawaz:
Closes #979 

Ok so basically:
1. Allow storage accesses everywhere with no restrictions. Only relevant in constraints though since everywhere else we expect a constant (?)
2. The flow is fairly simple middle end..
3. Changed the storage intrinsics to have separate `pre` and `post` versions.
4. In the backend, we have three possibilities when compiling a node:
    - All accesses are pre: insert the node with `Reads::Pre`
    - All accesses are post: insert the node with `Reads::Post`
    - The accesses are mixed: extract all the post accesses to their own nodes.


pintc/src/asm_gen/asm_builder.rs
```diff
@@ -86,15 +90,40 @@ impl<'a> AsmBuilder<'a> {
...
+        // memory for all of them. We do this ahead of time so that we know exactly how the size of
+        // memory allocated. Due to short-circuting, this also means that some allocations may not
+        // be used, but this is okay for now.
+        for access in &expr.collect_storage_accesses(contract) {
+            let num_keys = access.get_ty(contract).storage_keys(handler, contract)? as i64;
+            let access_size = access.get_ty(contract).size(handler, contract)? as i64;
+            let alloc_size = 2 * num_keys // for each key, add the data address then its length
+                + access_size; // then lay out the data read from each key, sequentiallly
+
+            asm.extend([PUSH(alloc_size), ALOC, POP]);
```
otrho, line 118: I think this ALOC might as well just be a single ALOC outside the loop, since we're throwing away the indices at runtime.  I noticed this recently when looking at map ASM gen.

pintc/tests/storage/bad_storage_accesses.pnt
```diff
@@ -96,50 +96,5 @@ predicate Test(
...
-// invalid position for accessing storage
-// @670..682: storage cannot be accessed in this position
-// storage can only be accessed in the initializer of a `let` declaration
-// invalid position for accessing storage
-// @175..187: storage cannot be accessed in this position
-// storage can only be accessed in the initializer of a `let` declaration
-// invalid position for accessing storage
-// @93..105: storage cannot be accessed in this position
-// storage can only be accessed in the initializer of a `let` declaration
 // compiler internal error: unexpected expression during compile-time evaluation
```
otrho, line 99: Should this test be a) renamed and/or b) pass now?  What is the error here?

otrho (commented):
I was going to make a lowering pass to split maps on expressions into a separate `let` to force it into a non-leaf node.  E.g. `(map x  in [1,2,3] { x+x})[1]` --> `let tmp = map x in [1,2,3] {x+x}; tmp[1]`.

But I think it makes sense to use the new `expr` node instead, and proper lowering can come later as an optimisation.

pintc/src/asm_gen/asm_builder.rs
mohammadfawaz: Fixed this.

pintc/tests/storage/bad_storage_accesses.pnt
mohammadfawaz: So, I renamed the test.. it still bad storage accesses that are used in places where constants are expected. We don't have good error messages for that so I'll open an issue to address that.

otrho: Ahh, OK.  Cool.

otrho (approved):

PR #1048, opened 2025-01-08
Optimize away identical `variable` and `constraint decls`

### LINKED ISSUE #1030: Optimize away identical vars and identical constraints
1. Two vars with identical initializers should be merged.
2. Two constraints with identical exprs should be merged.

Example:
```rust
predicate foo(x: int) {
    let a = x + 1;
    let b = x + 1;
    constraint a + b > 0;
    constraint a + b > 0;
}
```
should become something like:
```rust
predicate foo(x: int) {
    let a = x + 1;
    constraint a + a > 0;
}
```
To do this, we need a method to compare two exprs, ignoring spans. Then we'll have to do a search to figure out what can be removed. Also, when removing a var such as `b` above, we need to replace its uses with `a` (the other var we're merging with).

lethalgem:
Closes #1030


pintc/tests/optimizations/duplicate_variable_elimination.pnt
```diff
@@ -0,0 +1,934 @@
...
+    let c4 = w in 2..3; // should not be removed
+
+    // union values
+    let d4 = Fruit::Apples(w);
+    let e4 = Fruit::Apples(w);
+    let f4 = Fruit::Apples(x); // should not be removed
+
+    // generators - @mohammad, we can't have a variable represent a generator, only a constraint right?
+
+    // macro calls - uhhhh @mohammad, help please. Also, won't macros be gone by the time the optimizer runs?
```
lethalgem, line 185: @mohammadfawaz 

1. Can generators be used in a variable declaration? My guess is no.
2. Is there a way to use a macro call in a variable declaration?

pintc/tests/optimizations/duplicate_constraint_elimination.pnt
```diff
@@ -0,0 +1,621 @@
...
+    constraint w in 2..3; // should not be removed
+
+    // union values
+    constraint w3 == Fruit::Apples(w);
+    constraint w3 == Fruit::Apples(w);
+    constraint w3 == Fruit::Apples(x); // should not be removed
+
+    // generators - @mohammad, we can't have a variable represent a generator, only a constraint right?
+
+    // macro calls - uhhhh @mohammad, help please. Also, won't macros be gone by the time the optimizer runs?
```
lethalgem, line 195: Ditto for constraint decls.

pintc/src/expr.rs
```diff
@@ -372,6 +372,690 @@ impl Expr {
...
+                    call: rhs_call,
+                    path: rhs_path,
+                    ..
+                },
+            ) => {
+                todo!() // @mohammad, need help here please
+                        // if lhs_path != rhs_path {
+                        //     return false;
+                        // }
+            }
```
lethalgem, line 581: @mohammadfawaz :)

I'm not sure how to compare the call keys to each other. In nearly every instance in `transform.rs` we ignore any `MacroCall`s. My assumption is that we've already dealt with the Macros and we don't need to worry about them anymore by the time we're optimizing

pintc/tests/optimizations/duplicate_variable_elimination.pnt
mohammadfawaz: > Can generators be used in a variable declaration? My guess is no.

That is correct.

> Is there a way to use a macro call in a variable declaration?

Yes but macros are gone right after parsing so nothing to be done for them here.

pintc/tests/optimizations/duplicate_constraint_elimination.pnt
mohammadfawaz: Yeah same see my answer above.

pintc/src/expr.rs
mohammadfawaz: Yeah that is true.. However, for completeness, you can still compare the macro calls as you're suggesting (same macro name, same arguments, etc.)

pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -147,3 +151,170 @@ pub(crate) fn dead_select_elimination(contract: &mut Contract) {
...
+                                false
+                            }
+                        })
+                        .map(|_| {
+                            contract.exprs.insert(
+                                Expr::Path(original_var_key.get(pred).name.clone(), empty_span()),
+                                original_var_key.get_ty(pred).clone(),
+                            )
+                        });
+                }
```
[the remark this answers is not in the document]
mohammadfawaz: Yeah this is probably fine.. I will review properly when the PR is 100% ready.

pintc/tests/optimizations/duplicate_constraint_elimination.pnt
```diff
@@ -0,0 +1,621 @@
...
+    constraint 0 == f;
+    constraint 0 == f;
+    constraint 0 == g; // should not be removed
+
+    // external predicate calls - @mohammad why do we have a second address here? Is it because Foo would be the contract address and footest would be the address of the predicate in the contract?
+    let h = Foo@[addr0]::FooTest@[addr0](x, y);
+    let i = Foo@[addr0]::FooTest@[addr0](x, z);
+    constraint LocalTest@[](x, y) == h;
+    constraint LocalTest@[](x, y) == h;
+    constraint LocalTest@[](x, y) == h; // should not be removed
```
lethalgem, line 145: @mohammadfawaz 

why do we have a second address here? Is it because Foo would be the contract address and FooTest would be the address of the predicate in the contract?

pintc/tests/optimizations/duplicate_constraint_elimination.pnt
```diff
@@ -0,0 +1,621 @@
...
+
+    // ins
+    constraint w in [1, 2];
+    constraint w in [1, 2];
+    constraint x in [1, 2]; // should not be removed
+
+    // ranges - @mohammad, there's no way to assign a variable as a range right? This should test it properly, because the only inequity in the 'in' statement is the range
+    constraint w in 1..2;
+    constraint w in 1..2;
+    constraint w in 2..3; // should not be removed
```
lethalgem, line 186: @mohammadfawaz 

there's no way to assign a variable as a range right? This should test it properly, because the only inequity in the 'in' statement is the range, so it's still getting compared

pintc/tests/optimizations/duplicate_constraint_elimination.pnt
mohammadfawaz: Correct.. Just using the same variable `addr` here but that's likely never going to be the case.

pintc/tests/optimizations/duplicate_constraint_elimination.pnt
mohammadfawaz: Correct.. we can't assign a range to a variable.

pintc/src/asm_gen/tests.rs
```diff
@@ -513,16 +513,6 @@ fn binary_ops() {
                   Access(PredicateData)
                   Pred(Gt)
                 node 2
-                  Stack(Push(0))
```
mohammadfawaz, line 516: This unit tests seems to have
```
                constraint x > y;
                constraint x > y;
```
which is unnecessary. Let's remove one of them.

pintc/src/expr.rs
```diff
@@ -372,6 +372,675 @@ impl Expr {
...
+                        .get(contract)
+                        .eq(contract, rhs_elements[i].get(contract))
+                    {
+                        return false;
+                    }
+                }
+
+                lhs_range_expr
+                    .get(contract)
+                    .eq(contract, rhs_range_expr.get(contract))
```
mohammadfawaz, line 413: Maybe something like this?
```rust
lhs_elements.len() == rhs_elements.len()
    && lhs_elements.iter().enumerate().all(|(i, lhs_element)| {
        lhs_element.get(contract).eq(contract, rhs_elements[i].get(contract))
    })
    && lhs_range_expr
        .get(contract)
        .eq(contract, rhs_range_expr.get(contract))
```

And maybe the same can be done in other places.

pintc/src/expr.rs
```diff
@@ -372,6 +372,675 @@ impl Expr {
...
+                    ..
+                },
+            ) => {
+                lhs_union_expr
+                    .get(contract)
+                    .eq(contract, rhs_union_expr.get(contract))
+                    && lhs_variant_ty.eq(contract, rhs_variant_ty)
+            }
+
+            _ => false,
```
mohammadfawaz, line 1040: It's better to just list all remaining exprs.. more robust when new exprs are added in the future.

pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -147,3 +151,170 @@ pub(crate) fn dead_select_elimination(contract: &mut Contract) {
...
+                        }
+                    });
+                }
+            }
+        }
+
+        // replace all uses of the duplicate var exprs with the originals
+        for (original_expr_key, dupe_expr_key) in dupe_var_exprs {
+            contract.replace_exprs(Some(pred_key), dupe_expr_key, original_expr_key);
+        }
```
mohammadfawaz, line 263: Is this necessary? If we have two identical variables:
```
let a = 5;
let b = 5;
constraint a + b == 10;
```
can't we just delete `b` (as we're doing later) and just replace the path to `b` to a path to `a` in the constraint? No need to update the expr of the var itself since it's going to be deleted anyways. Or am I missing another case here?

pintc/tests/optimizations/duplicate_variable_elimination.pnt
```diff
@@ -0,0 +1,930 @@
...
+}
+
+predicate LocalTest(local_x: int, local_y: bool) {}
+
+predicate Test(w: int, x: int, y: bool, z: bool, w2: Coin, x2: Coin) {
+    let addr0 = 0x0000000000000000000000000000000000000000000000000000000000000000;
+
+    // binary ops
+    let a = x + 1;
+    let b = 1 + x;
```
mohammadfawaz, line 29: Is there an example with 3 or more variables being collapsed into 1? Same with constraints.

mohammadfawaz:
So, finding duplicates is an interesting problem. What you've done is of course `O(n^2)` because you have nested loops. There are generally better approaches such as https://www.geeksforgeeks.org/find-duplicates-in-on-time-and-constant-extra-space/ - do you want to see if this can be applied here in some way? It would be exactly the same because hashing won't work directly (because of spans) so you'll need to figure something else out. The point is to do a first pass in `O(n)`, collect frequency information, then act upon it.

pintc/src/expr.rs
lethalgem: https://github.com/essential-contributions/pint/blob/c0e93b60203b6857b6769ceb95345f0274353f39/pintc/src/expr.rs#L954-L978

Since we're matching on a tuple of the possible expr combinations, I've opted to just check the first position. This should let us know that we need to make modifications if we add/remove exprs in the future

pintc/src/expr.rs
```diff
@@ -372,6 +372,613 @@ impl Expr {
...
+    pub fn eq(&self, contract: &Contract, other: &Self) -> bool {
+        match (self, other) {
+            (
+                Expr::Immediate {
+                    value: lhs_value, ..
+                },
+                Expr::Immediate {
+                    value: rhs_value, ..
+                },
+            ) => lhs_value == rhs_value,
```
sezna, line 384: I think there are generally two motivations to write a new function. One is code reuse: keeping things DRY. The other is readability. I can tell what this huge clause is doing, but I'd rather it read something like:
```rust
match (self, other) {
  use Expr::*;
   (Immediate { value: lhs, .. }, Immediate { value: rhs, .. }) => immediate_eq(lhs, rhs),
   (Array { elements: lhs_elements, range: lhs_range, .. }, Array { elements: lhs_elements, range: rhs_elements, ..) => array_eq(lhs_elements, lhs_range, rhs_elements, rhs_range)
    ...etc
}
```

Rust unfortunately doesn't support enum values, so another more invasive way to clean this up even more would be to wrap the enum variants' contents in their own structs. Then, it'd be even more readable and clean:

```rust
enum Expr {
    Array(Array),
    Immediate(Immediate),
    ...etc
}
match (self, other) {
  use Expr::*;
   (Immediate(lhs), Immediate (rhs)) => immediate_eq(lhs, rhs),
   (Array(lhs), Array(rhs))          => array_eq(lhs, rhs)
   ...etc
```

sezna: The second method is potentially undesirably invasive depending on how much `Expr` refactoring you want to do

pintc/tests/optimizations/duplicate_variable_elimination.pnt
lethalgem: https://github.com/essential-contributions/pint/blob/a4548ac61224adf86e066b7c748f6351a8e44665/pintc/tests/optimizations/duplicate_constraint_elimination.pnt#L193-L197

pintc/src/predicate/optimize/dead_code_elimination.rs
lethalgem: Good catch! Definitely doing unnecessary work here

pintc/src/expr.rs
```diff
@@ -1103,3 +895,476 @@ impl Expr {
...
+                .get(contract)
+                .eq(contract, rhs_conditions[i].get(contract))
+        })
+        && lhs_body == rhs_body
+}
+
+pub fn map_eq(
+    contract: &Contract,
+    lhs_param: &Ident,
+    lhs_range: &ExprKey,
```
sezna, line 1336: I could be wrong but it _looks_ like slotmap keys are `Copy` by default: https://docs.rs/slotmap/latest/src/slotmap/lib.rs.html#456 (and https://docs.rs/slotmap/latest/src/slotmap/lib.rs.html#260)

perhaps no `&` is needed here?

otrho:
> It would be exactly the same because hashing won't work directly (because of spans) so you'll need to figure something else out. The point is to do a first pass in `O(n)`, collect frequency information, then act upon it.

That page is specifically sorting numbers from 0 to n-1 which makes it a bit easier -- I think for complex data types hashing is the only option, though to only hash the parts which we're comparing, i.e., not the spans.

pintc/src/expr.rs
otrho: Yep, they're definitely (and very handily) `Copy.`

pintc/src/expr.rs
otrho: Yep, the new functions version is much easier to read, and ideally it'd be some sort of special equivalency trait (or maybe just `PartialEq`) that we'd use, but that would require wrapping  like the second method too.

At this stage `Expr` is too widely used to be mucking around with it that much. :) Definitely an option for the almighty Version 2... in fact the design proposal I have in mind for Version 2 already has all the nodes wrapped for better type safety along the pipeline.

pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -147,3 +151,169 @@ pub(crate) fn dead_select_elimination(contract: &mut Contract) {
...
+                .iter()
+                .any(|(original_var_key, dupe_var_key)| {
+                    original_var_key == var_key || dupe_var_key == var_key
+                })
+            {
+                continue;
+            };
+
+            for (subsequent_var_key, subsequent_var) in pred_variables.iter().skip(i + 1) {
+                if let Some(pred) = contract.preds.get(pred_key) {
```
mohammadfawaz, line 193: So you're checking if `contract.preds.get(pred_key)` exists multiple times in this file. If you want, you can do this check at the beginning only once (right after `for pred_key in contract.preds.keys().collect::<Vec<_>>() {`)

pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -147,3 +151,169 @@ pub(crate) fn dead_select_elimination(contract: &mut Contract) {
...
+                }
+
+                if var
+                    .expr
+                    .get(contract)
+                    .eq(contract, subsequent_var.expr.get(contract))
+                {
+                    dupe_var_decls.push((*var_key, *subsequent_var_key));
+                }
+            }
```
mohammadfawaz, line 209: These two checks can be combined due to short-circuiting:
```rust
if var_key.get_ty(pred).eq(contract, subsequent_var_key.get_ty(pred))
&& var.exp.get(contract).eq(contract, subsequent_var.expr.get(contract)) { 
    ..
}
```

pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -147,3 +151,169 @@ pub(crate) fn dead_select_elimination(contract: &mut Contract) {
...
+                    if let Expr::Path(name, _) = expr.get(contract) {
+                        *name == original_var_key.get(pred).name
+                    } else {
+                        false
+                    }
+                });
+
+                // if the original variable is unused, check if the duplicate is in use and needs to be replaced
+                if original_path_expr.is_none() {
+                    original_path_expr = contract
```
mohammadfawaz, line 231: So this is very confusing since you're changing the "meaning" of original here. First thing I would do is to merge this block of code with the `if let` later as in:
```
if let Some(original_path_expr) = original_path_expr {
    // handle original
} else {
    // handle dup
}
```
and we'll see where this goes

pintc/src/predicate/optimize/dead_code_elimination.rs
```diff
@@ -147,3 +151,169 @@ pub(crate) fn dead_select_elimination(contract: &mut Contract) {
...
+/// constraint a == x + 1;
+/// constraint a == x + 1;
+/// constraint a == b + 1;
+/// becomes
+/// constraint a == x + 1;
+/// constraint a == b + 1;
+pub(crate) fn duplicate_constraint_elimination(contract: &mut Contract) {
+    for pred_key in contract.preds.keys().collect::<Vec<_>>() {
+        if let Some(pred) = contract.preds.get(pred_key) {
+            let mut redundant_constraint_locs: Vec<usize> = vec![];
```
mohammadfawaz, line 288: Maybe changed `locs` to `indices` or something.

mohammadfawaz:
I just realized that `And` and `Or` are not currently commutative due to short-circuiting. Consider this:
```rust
let x = storage::x != nil && storage::x' == storage::x + 1
```
It's technically illegal to change this to 
```
let x =  storage::x' == storage::x + 1 && storage::x != nil
```
because in the first version, the RHS will only be executed if the LHS is `true` and so, if `storage::x == nil` the VM won't panic. In the second version, if `storage::x == nil`, then the VM will panic.

So, what I suggest for now is to not consider `AND` and `OR` commutative.

pintc/src/predicate/optimize/dead_code_elimination.rs
lethalgem: Here is the option with replaced blocks:
```rust
// collect all of the duplicate var exprs and duplicate paths to be cleared out
let mut dupe_var_exprs: Vec<(ExprKey, ExprKey)> = vec![];
let mut dupe_paths: Vec<(ExprKey, ExprKey)> = vec![];
for (original_var_key, dupe_var_key) in &dupe_var_decls {
    let original_expr_key = original_var_key.get(pred).expr;
    let dupe_expr_key = dupe_var_key.get(pred).expr;
    dupe_var_exprs.push((original_expr_key, dupe_expr_key));

    // get path expr if original exists, if not try to get one of it's dupes, and if that fails, skip the process
    let original_path_expr = contract.exprs(pred_key).find(|expr| {
        if let Expr::Path(name, _) = expr.get(contract) {
            *name == original_var_key.get(pred).name
        } else {
            false
        }
    });

    if let Some(original_path_expr) = original_path_expr {
        contract.exprs(pred_key).for_each(|expr| {
            if let Expr::Path(name, _) = expr.get(contract) {
                if *name == dupe_var_key.get(pred).name {
                    dupe_paths.push((original_path_expr, expr))
                }
            }
        });
    } else {
        if let Some(path_expr) = contract
            .exprs(pred_key)
            .find(|expr| {
                if let Expr::Path(name, _) = expr.get(contract) {
                    *name == dupe_var_key.get(pred).name
                } else {
                    false
                }
            })
            .map(|_| {
                contract.exprs.insert(
                    Expr::Path(original_var_key.get(pred).name.clone(), empty_span()),
                    original_var_key.get_ty(pred).clone(),
                )
            })
        {
            contract.exprs(pred_key).for_each(|expr| {
                if let Expr::Path(name, _) = expr.get(contract) {
                    if *name == dupe_var_key.get(pred).name {
                        dupe_paths.push((path_expr, expr))
                    }
                }
            });
        }
    }
}
```

Here is an alternative that just removes the need to duplicate the code for pushing the keys into the dupe list
```rust
// collect all of the duplicate var exprs and duplicate paths to be cleared out
let mut dupe_var_exprs: Vec<(ExprKey, ExprKey)> = vec![];
let mut dupe_paths: Vec<(ExprKey, ExprKey)> = vec![];
for (original_var_key, dupe_var_key) in &dupe_var_decls {
    let original_expr_key = original_var_key.get(pred).expr;
    let dupe_expr_key = dupe_var_key.get(pred).expr;
    dupe_var_exprs.push((original_expr_key, dupe_expr_key));

    let path_expr = if let Some(original_path_expr) =
        contract.exprs(pred_key).find(|expr| {
            if let Expr::Path(name, _) = expr.get(contract) {
                *name == original_var_key.get(pred).name
            } else {
                false
            }
        }) {
        Some(original_path_expr)
    } else {
        // if the original variable is unused, check if the duplicate is in use and needs to be replaced
        if let Some(new_path_expr) = contract
            .exprs(pred_key)
            .find(|expr| {
                if let Expr::Path(name, _) = expr.get(contract) {
                    *name == dupe_var_key.get(pred).name
                } else {
                    false
                }
            })
            .map(|_| {
                contract.exprs.insert(
                    Expr::Path(original_var_key.get(pred).name.clone(), empty_span()),
                    original_var_key.get_ty(pred).clone(),
                )
            })
        {
            Some(new_path_expr)
        } else {
            None
        }
    };

    if let Some(path_expr) = path_expr {
        contract.exprs(pred_key).for_each(|expr| {
            if let Expr::Path(name, _) = expr.get(contract) {
                if *name == dupe_var_key.get(pred).name {
                    dupe_paths.push((path_expr, expr))
                }
            }
        });
    }
}
```

Neither are super legible. I could create a couple of local functions that would remove the need for duplicating pushing the keys to the list or searching for a path expr in use. Though both of those options may make it less obvious what is actually going on here.

ex. (also in the latest commit)
```rust
fn collect_all_duplicate_paths(
contract: &Contract,
pred_key: PredKey,
dupe_paths: &mut Vec<(ExprKey, ExprKey)>,
dupe_var_key: VariableKey,
path_expr: ExprKey,
pred: &Predicate,
) {
contract.exprs(pred_key).for_each(|expr| {
    if let Expr::Path(name, _) = expr.get(contract) {
        if *name == dupe_var_key.get(pred).name {
            dupe_paths.push((path_expr, expr))
        }
    }
});
}
```

What I want to get across to a future reader is that we
1. check to see if the original var is in use (has a path)
2. if not we check if any duplicate var decl is used (has a path), in which case we create a path for the original var and then use that to replace any duplicate paths
3. if there are no uses of the original or duplicate var decls, then we skip any replacements

While it is true that mutating `original_path_expr` does change it's meaning, it's purpose remains the same. We need a path that matches the original expr so that we can replace all other duplicate paths

pintc/src/expr.rs
```diff
@@ -522,3 +923,474 @@ impl Expr {
...
+            }
+
+            ((IntrinsicKind::Internal(lhs_kind), _), (IntrinsicKind::Internal(rhs_kind), _)) => {
+                lhs_kind == rhs_kind
+            }
+
+            ((IntrinsicKind::Error, _), (IntrinsicKind::Error, _)) => true,
+
+            _ => false,
+        }
```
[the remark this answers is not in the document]
lethalgem: Understood. We don't have any information on what the error actually is here. Do we want to bring in the `handler` and go diving through it to find what the errors are? Or do we want to skip this altogether and say an error means it isn't equivalent?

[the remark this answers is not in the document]
mohammadfawaz: Meh.. not too important. We shouldn't have errors when we do this anyways. `Error` here is just a fallback option and only relevant for error recovery.

mohammadfawaz (approved):
Awesome, thanks for addressing all the comments.

PR #1055, opened 2025-01-16
Improve error if a `const` is used in a `predicate`

### LINKED ISSUE #1052: Improve `const` error when used in a `predicate`
`const` is not allowed inside of a predicate. This is error should be reported to the user.

```pnt
storage {
    x: int,
}

predicate Test() {
    const a = storage::x;
}
```

reports the error

```zsh
Error: expected `::`, `an identifier`, `constraint`, `if`, `let`, `macro_name`, `match`, `use`, or `}`, found `const`
   ╭─[test.pnt:6:5]
   │
 6 │     const a = storage::x;
   │     ──┬──  
   │       ╰──── expected `::`, `an identifier`, `constraint`, `if`, `let`, `macro_name`, `match`, `use`, or `}`
───╯
```

lethalgem:
Closes #1052

### DIFF
--- a/pintc/src/pint_parser.lalrpop
+++ b/pintc/src/pint_parser.lalrpop
@@ -42,6 +42,15 @@ Decl: () = {
 
 // Decls allowed inside predicates.
 PredicateInnerDecl: () = {
+    <l:@L> ConstDecl ";" <r:@R> => {
+        let span = (context.span_from)(l, r);
+        handler.emit_err(Error::Parse {
+            error: ParseError::UnsupportedConstLocation {
+                span: span.clone(),
+            },
+        });
+    },
+
     ConstraintDecl ";",
     IfDecl,
     MatchDecl,
--- a/pintc/src/error/parse_error.rs
+++ b/pintc/src/error/parse_error.rs
@@ -69,6 +69,8 @@ pub enum ParseError {
     TypeNotSupported { ty: String, span: Span },
     #[error("Unsupported literal")]
     LiteralNotSupported { kind: String, span: Span },
+    #[error("`consts` must be declared outside of a `predicate`")]
+    UnsupportedConstLocation { span: Span },
 }
 
 impl ReportableError for ParseError {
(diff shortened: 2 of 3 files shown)


pintc/src/pint_parser.lalrpop
```diff
@@ -42,6 +42,15 @@ Decl: () = {
...
 // Decls allowed inside predicates.
 PredicateInnerDecl: () = {
+    <l:@L> ConstDecl ";" <r:@R> => {
+        let span = (context.span_from)(l, r);
+        handler.emit_err(Error::Parse {
+            error: ParseError::UnsupportedConstLocation {
+                span: span.clone(),
+            },
+        });
+    },
```
otrho, line 52: I don't really like this, sorry.  But at the same time I don't know a better way off the top of my head.

I can see that by not having `ConstDecl` in the `PredicateInnerDecl` we will just get a parse error, of whatever it is -- "unexpected 'const', expecting blah..." -- which isn't helpful enough, and the point of this change.

But deliberately matching a rule so that we can emit a specific error isn't really what the parser should be doing.  If possible it'd be better to have a step which analyses parse errors and will convert them into better errors.  But, hrmm, how do we know that _this_ parse error which starts with a `const` token is actually declared within a predicate?  We can't, at least with this parser, and TBH any decent parser.   (The job of the parser is to convert a string of characters or tokens into an AST and not to be making semantic analyses. Unfortunately our parser has overstepped this boundary in a few places, mostly to avoid over-engineering generating a `Contract` in the early days, but now in hindsight has just made it too complex.)

So... a compromise would be to allow `ConstDecl` in the rule and then the AST, and to check it later, find it misplaced and emit this error.  But we don't create an AST.  Argh.  All we could do is push the error into a special spot in the context and report it later, which isn't particularly helpful.

Dammit, so this comment is now just a rant about how the parser is broken (which isn't news) and this change isn't great but it truly seems to be the only way to emit this error.  😒

otrho (approved):

pintc/src/pint_parser.lalrpop
mohammadfawaz: We could just add a list of consts too each `Predicate` object and then error out on each element in the list 🤷 but that looks like an anti-pattern.

otrho: Yeah, exactly.

PR #1063, opened 2025-02-10
Improve `const` evaluator error

### LINKED ISSUE #1049: Improve const evaluator error
In many cases, (e.g. storage accesses), the const evaluator will fail and we should emit a proper error. Currently, we always emit an internal error which is not great. If a storage accesses is used in a place where a compile-time constant is expected, a clear error message should be displayed.

lethalgem:
Closes #1049


lethalgem:
@mohammadfawaz Still in draft, got lots of tests to update and need to double check where I'm forcing errors to be reported to the user vs where we're sticking to internal errors.

I wanted to get this draft up for you to take a look at. It solves the issue, but I think it adds complexity. We're basically flagging whether or not we want the final error from the evaluator to be internal or not. It means the evaluator is pulling double duty. Also I've given it a helper function inside the evaluator, but perhaps it belongs in the handler instead?

Happy to get your thoughts on whether or not this is satisfactory and can be improved.

lethalgem:
Notes after talk with @mohammadfawaz 

Three levels to this
1. The user did something where the code is valid, but can't work. Ex. `const A = storage::x`. They need to be told that this can't be evaluated to a const
2. Best effort. Ex. inside of the optimization phase. try to evaluate as a const, it fails, move on. It's not an error
3. When you expect the evaluator to pass. If not, then it's a bug. And should be reported as such as a compiler error.

Be sure to check out the scenario where one side of a binary op cannot be evaluated to a const. Where do we fail? Does the user get the proper error?

pintc/src/expr/evaluate.rs
```diff
@@ -525,13 +525,16 @@ impl Evaluator {
...
             Expr::Error(_)
-            | Expr::LocalStorageAccess { .. }
-            | Expr::ExternalStorageAccess { .. }
-            | Expr::MacroCall { .. }
             | Expr::IntrinsicCall { .. }
-            | Expr::LocalPredicateCall { .. }
-            | Expr::ExternalPredicateCall { .. }
             | Expr::Range { .. }
             | Expr::Generator { .. }
             | Expr::Map { .. }
```
lethalgem, line 540: Thinking more on it, it appears that most of these (if not all) expressions should lead to a user error that states "invalid const". Is that correct?

An `intrinsic` call probably requires more nuance because `__size_of` could be a `const`. Other calls definitely could not be `const`.

A `map` can't be `const` afaik because it can only be used in storage.

A `generator`, much like a `macrocall` should have been dealt with already before the `evaluator` is used. Actually I guess both `generator` and `macrocall` should be internal compiler errors instead of user errors.

Are `ranges` a `const` in pint? I'm not actually sure where a `range` is valid other than in a `generator`.

mohammadfawaz: Yeah pretty much all of these should be a user error (not internal) except for:
`Expr::Error` and `MacroCall` because seeing either of these at this point is unexpected and should be an internal error. The others are fine in principle but we're generally unable to evaluate them at compile time hence the user error.

pintc/src/error/compile_error.rs
```diff
@@ -383,6 +383,8 @@ pub enum CompileError {
     },
     #[error("invalid map range type")]
     InvalidMapRangeType { found_ty: String, span: Span },
+    #[error("invalid constant")]
+    InvalidConst { span: Span },
```
mohammadfawaz, line 387: Maybe something closer to Rust:
image

lethalgem:
There is a test for every `Expr` except `Expr::Range` and `Expr::Generator`. Neither can be used as constants, as dictated by the parser.

otrho:
Unfortunately we were too slow with this one and now it has conflicts.  Sorry @lethalgem

pintc/src/expr/evaluate.rs
```diff
@@ -515,22 +515,24 @@ impl Evaluator {
...
+            | Expr::Match { span, .. }
+            | Expr::IntrinsicCall { span, .. }
+            | Expr::Range { span, .. }
+            | Expr::Generator { span, .. }
+            | Expr::Map { span, .. }
+            | Expr::Nil(span) => Err(handler.emit_err(Error::Compile {
+                error: CompileError::InvalidConst { span: span.clone() },
+            })),
+
+            Expr::Error(_) | Expr::MacroCall { .. } | Expr::AsmBlock { .. } => Err(handler
```
lethalgem, line 531: @mohammadfawaz I left `AsmBlocks` as an error. Our parser doesn't allow it to be used as an expression for a const, so it should be a compiler bug if we ever find an `AsmBlock` as a `const`.

mohammadfawaz (approved):

PR #1064, opened 2025-02-11
[WIP] Type `Optional`

mohammadfawaz:
Very much a wip


pintc/src/asm_gen/asm_builder.rs
```diff
@@ -312,7 +318,19 @@ impl<'a> AsmBuilder<'a> {
...
-                Immediate::Error | Immediate::Nil | Immediate::Real(_) | Immediate::String(_) => {
+                Immediate::Optional { value, value_size } => {
+                    if let Some(value) = value {
+                        let value_size = compile_immediate(handler, asm, value)?;
+                        asm.push(PUSH(1));
+                        Ok(1 + value_size)
+                    } else {
+                        for _ in 0..1 + value_size {
+                            asm.push(PUSH(0));
+                        }
```
otrho, line 329: I think this could be a `RES`.  I assume it works when the stack is non-empty.

pintc/src/asm_gen/asm_builder.rs
```diff
@@ -521,6 +542,78 @@ impl<'a> AsmBuilder<'a> {
                 asm.push(SUB);
                 Ok(Location::Stack)
             }
+
+            UnaryOp::Unwrap => {
```
otrho, line 546: IMO, this is too big and needs to go in a method.

pintc/src/asm_gen/asm_builder.rs
```diff
@@ -1088,6 +1109,28 @@ impl<'a> AsmBuilder<'a> {
...
+    ) -> Result<Location, ErrorEmitted> {
+        let expr_ty = expr.get_ty(contract);
+        if let Some(value) = value {
+            self.compile_expr(handler, asm, value, contract, pred)?;
+            asm.push(PUSH(1)); // tag
+        } else {
+            // tag and value are both zeros
+            for _ in 0..expr_ty.size(handler, contract)? {
+                asm.push(PUSH(0));
+            }
```
otrho, line 1129: Also candidate for `RES`?

pintc/src/predicate/analyse/type_check/validate.rs
```diff
@@ -236,4 +237,47 @@ impl Contract {
...
+        &mut self,
+        handler: &Handler,
+    ) -> Result<(), ErrorEmitted> {
+        fn check_any_type(ty: &Type, handler: &Handler) {
+            match ty {
+                Type::Any(span) => {
+                    handler.emit_err(Error::Compile {
+                        error: CompileError::UninferrableType { span: span.clone() },
+                    });
+                }
```
otrho, line 251: Do you have an example of this happening?  It'd be way better if we could just catch these in the type checker rather than in a validation pass.

pintc/src/predicate/analyse/type_check/check_exprs.rs
```diff
@@ -1388,11 +1468,25 @@ impl Contract {
...
         }
 
-        let ary_ty = array_expr_key.get_ty(self);
-        if ary_ty.is_unknown() {
+        let ty = array_expr_key.get_ty(self);
+        if ty.is_unknown() {
             return Inference::Dependant(array_expr_key);
         }
 
+        let is_storage_access = array_expr_key.is_storage_access(self);
```
otrho, line 1476: All of the boilerplate that this adds is unfortunate.  I feel like one day we should refactor it out to be a recursive call or something, if possible.

pintc/src/expr.rs
```diff
@@ -196,6 +199,10 @@ pub enum Immediate {
         value: Option<Box<Immediate>>,
         decl: UnionKey,
     },
+    Optional {
+        value: Option<Box<Immediate>>,
+        value_size: usize,
+    },
 }
```
otrho, line 206: Is there a specific reason for adding an `Expr::Optional` _and_ a `Immediate::Optional`?  `Expr::Optional` isn't enough alone?

otrho (commented):
This is great, thanks.  A bunch of comments though...

Also, I'm not convinced we need `val(..)`.  `nil` is fine for checking `x == nil` (although we could achieve the same with an `is_nil(x)` intrinsic and not need `nil`), but I don't see `val(..)` as useful since we have `!` unwrapping.

pintc/src/predicate/analyse/type_check/validate.rs
otrho: I understand if we're comparing `nil` to `nil` then we can't infer a type... but then the type should just be `Nil`?  Otherwise, if there is any use of the storage value at all then it should infer that type.

pintc/src/predicate/analyse/type_check/check_exprs.rs
mohammadfawaz: Created an issue.

pintc/src/expr.rs
mohammadfawaz: Remove `Expr::Optional` altogether in favour `Expr::Nil`. That's because I also removed `val` so there's only one possibility which is `nil`.

pintc/src/predicate/analyse/type_check/validate.rs
mohammadfawaz: Alright so yeah switched the type to `Nil` instead of `Any` in this case, but we still can't claim that `Nil == Nil` because we don't have size information. Right now, there are other cases where the type of `Nil` is not inferable such as `constraint `[ storage::a, storage::b ] == [ nil, nil ]` but that's something we could fix.

pintc/tests/basic_tests/nil.pnt
```diff
@@ -51,28 +51,13 @@ predicate test() {
...
-//     constraint (__size_of(::z) != 0);
-//     constraint ((__size_of(::z) == 0) && (__size_of(::x) != 0));
-//     constraint ((__size_of(::z) != 0) && (__size_of(::x) == 0));
-//     constraint __eq_set(__mut_keys(), {0});
-// }
+// typecheck_failure <<<
+// uninferrable type
+// @166..169: type of this expression cannot be inferred
+// uninferrable type
+// @159..162: type of this expression cannot be inferred
```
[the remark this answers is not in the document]
mohammadfawaz: See the other comment.

mohammadfawaz:
> although we could achieve the same with an `is_nil(x)`

While the intrinsic might be easier, something about it doesn't feel right. `is_nil` would actually be more efficient though since we won't have to do large compares.. we could just check the tag (though we could optimize things ourselves). But yeah I'm conflicted. I like that `nil` is just a value here like any other, waiting to have its type inferred, so someone could write:
```
storage {
    a: int[3]
}

constraint storage::a = [ 5, nil, 3 ]
```

pintc/src/predicate/analyse/type_check/validate.rs
otrho: OK.  Since the type system is actually pretty simple and we're not doing 'difficult' type inference, it would seem a bit lame for use to give up and say we couldn't infer the type.  Perhaps for very specifc `Nil` related errors we can have Errors which explain exactly why the type is unknown, and even hint how to fix it.

otrho:
> I like that `nil` is just a value here like any other, waiting to have its type inferred

I agree, now that we might be passing optional values to predicates, we'd need to use `nil` there too anyway.

otrho (approved):
:shipit:

Issue #987, opened 2024-11-14
Change jump indices to labels and resolve them later


mohammadfawaz:
This helps adding peephole optimizations

PR #1066, opened 2025-02-18
Add ASM for maps over arrays of arrays.

otrho:
This also includes a fix for a bug which didn't quite generate the correct ASM for `let a = storage::b + storage::b';`.

The case where array expressions/immediates need to be copied from the stack to memory has also been improved.

Previously I had it making a full copy of the stack entries, saving them to memory and then popping all the originals.  Horribly inefficient.  The better solution in this PR is to instead compile the array generating ASM separately and if it's on the stack then prefix it with the buffer address before adding it to the 'real' ASM.  (This is all due to `STOR` requiring the address before the data, which has an issue to change.)

This will fail if the array generation requires jumps to addresses.  They'll be wrong and combining them will create erroneous jumps.  This won't be an issue once we add an IR with labels for jumps, i.e,  #987.


mohammadfawaz (approved):
Awesome, thanks!

PR #1076, opened 2025-02-25
Add asm blocks as possible var initializers

mohammadfawaz:
Add `asm` blocks as expressions but only as initializers for `let` variables (for now). Example:
```
    let y = 5555;
    let z = 3333;
    let w = 2222;

    let sum: int = asm(w, y, z) {
        PUSH 0
        LOD // load w
        PUSH 1
        LOD // load y
        ADD
        PUSH 2
        LOD // load z
        ADD
    };
```
A few things:
- The type of an `asm` block is determined based on the type of the variable. If the variable does not have a user-specified type, then we will error out.
- An `asm` block accepts a list of arguments that are arbitrary expressions. These argument are laid out in order at the start of the memory for he compute node representing the variable initializer.
- The result of an `asm` block is read directly from the stack as the type of the `asm` block. Hopefully things match, otherwise bad things will happen (panics etc.)
- `asm` blocks can have anything in them including jumps and repeats. That's because the whole asm block is an independent program.
- There is some boilerplate code that is added to an asm block to store its result back in memory where the value of the variable will now live.


pintc/src/pint_parser.lalrpop
```diff
@@ -1238,6 +1247,47 @@ IdentFromToken<Tok>: Ident = {
...
+                ops,
+                span: span.clone(),
+            },
+            Type::Unknown(span),
+        )
+    }
+}
+
+AsmOp: AsmOp = {
+    <l1:@L> <op:Ident> <l2:@L> <arg:"int_lit"?> <r:@R> => {
```
otrho, line 1268: It looks like you're not using `l1` here.  And we should have it so parse_int_immediate() can't fail, or just panics.  By parsing the string of the int we should be validating that it's correct before we try to convert it to a number.  Can't fail.

tests/validation_tests/asm.pnt
```diff
@@ -0,0 +1,67 @@
...
+    let sum: int = asm(w, y, z) {
+        PUSH 0
+        LOD // load w
+        PUSH 1
+        LOD // load y
+        ADD
+        PUSH 2
+        LOD // load z
+        ADD
+    };
```
otrho, line 25: Shouldn't `w`, `y` and `z` have prefix sizes?  I thought `w`s length would be at `0`, and it would be 1.  So `w` at 1, `y` at 3 and `z` at 5?

pintc/src/pint_parser.lalrpop
```diff
@@ -227,6 +227,15 @@ LetDecl: () = {
             .insert_variable(handler, mod_prefix, name.1, &name.0, ty, init, span)
             .map(|_| ());
     },
+    <l:@L> "let" <name:LetName> <ty:(":" <Type>)?> "=" <init:AsmBlockExpr> <r:@R> => {
```
otrho, line 230: For `asm` blocks the type shouldn't be optional.

otrho (commented):
I've left a few comments but there's only one which I think needs to be changed.  ASM blocks are now the only expression we have which don't have an inferable type.  But we're strict and only allow them on the RHS of `let`s so those `let`s must be typed.  Then there can't be any ambiguity -- the ASM block is given its type by the declaration.

pintc/src/pint_parser.lalrpop
otrho: I wonder without actually trying if we have mandatory types for ASM `let`s then we could perhaps type them with a quick pass in the type checker and then avoid the trickiness added in this PR for inference and the ASM comment in #1079.

tests/validation_tests/asm.pnt
otrho: I've been informed that the prefix sizes have been removed, since they're pretty much redundant.

otrho:
> Some general feedback after using this today. I think it would be a lot easier to use if it didn't try to store you memory for you. It should just set your memory from the input programs and then let you do whatever you want. If you mess up the output then that's on the user not the compiler. So a fresh empty stack and being able to choose what to output without that needing to be whatever is on the stack would be better IMHO

The compiler will guarantee that the size of the result of the node will be correct (although perhaps at the moment if the result is too small it might not pad out the result, but it should).  This in turn means that the weirdness from corrupt data shouldn't affect other nodes.  If an ASM block leaves memory the wrong size for its type then it'll affect all the other parent nodes for a leaf and working out what the problem is will be much harder.

Building complex values on the stack should be fairly easy/identical to building them in memory by using `RES` and `STOS`.

freesig:
Imagine the scenario where you want to reserve some stack for known memory locations. Firstly you have to know that there's a push 0 inserted here. Then after you have done your work you have no way of clearing the reserved stack because it's below what you need to leave on the stack to return. So it gets returned as memory. I reckon just leave the guard rails off. Its already really unsafe to use asm so there no real reason to add this one thing

pintc/src/asm_gen/asm_builder.rs
```diff
@@ -449,6 +464,98 @@ impl<'a> AsmBuilder<'a> {
         }
     }
 
+    /// Given an assembly block, convert each instruction in it into its corresponding op in
+    /// `essential_asm::short` and append it to `asm`.
+    ///
+    /// TODO: we should really auto-generate this from the `essential-asm-spec` crate using a proc
+    /// macro
```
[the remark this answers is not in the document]
mohammadfawaz: Yeah keeping the issue open but low priority.

pintc/src/expr.rs
```diff
@@ -158,6 +163,12 @@ impl Default for Ident {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct AsmOp {
+    pub(super) op: Ident,
+    pub(super) arg: Option<i64>,
+}
+
```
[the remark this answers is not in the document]
mohammadfawaz: Also, if we do this, then we would have to error out in the parser if someone writes `NOTPUSH 2`. If one day, we have separate trees in the compiler, then we could do all kinds of stuff: parse tree, typed tree, flattened tree, etc.

pintc/src/pint_parser.lalrpop
mohammadfawaz: Wait, I'm confused. The lexer already checks that the string is an integer. `parse_int_immediate` then checks whether it's a _valid_ int. In this case, we also want to exclude b256s. So, `parse_int_immediate` will basically fail if the int is too wide for an `i64`.

mohammadfawaz: I reworded the error message a bit. If the literal is not an integer at all, the parser will emit the usual `expected...found` error.

tests/validation_tests/asm.pnt
mohammadfawaz: I'll still open an issue to discuss whether sizes are needed or not. They may become relevant when dynamic arrays are added.

pintc/src/pint_parser.lalrpop
mohammadfawaz: Done.

mohammadfawaz:
> Imagine the scenario where you want to reserve some stack for known memory locations. Firstly you have to know that there's a push 0 inserted here. Then after you have done your work you have no way of clearing the reserved stack because it's below what you need to leave on the stack to return. So it gets returned as memory. I reckon just leave the guard rails off. Its already really unsafe to use asm so there no real reason to add this one thing

The reason why we need to push 0 on the stack before anything else is because `STO` and `STOR` expect the memory index _before_ the data. If we change the order as suggested by @otrho in https://github.com/essential-contributions/essential-base/issues/247, then all the assembly added by the compiler would then show up after the `asm` block content. So, from the point of view of the user, it should be seamless. Would that be better? I don't mind removing the boilerplate code but I do like that it keeps things a bit more sane.

freesig:
Even if the order is changed then you still end up with the issue of not being able to remove the reserved scratch space without shifting everything to memory, popping and then pulling back to the stack and then the boiler plate sends it back to memory. That's a lot of clones if we are talking about 100,000 transfers in memory

pintc/src/expr.rs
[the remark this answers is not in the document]
otrho: `NOTPUSH 2`  would still parse, and would be translated into `NOTPUSH, PUSH(2)`.  Most of the stack languages I've seen will just implicitly push any immediates found in the input.

The parser shouldn't be doing semantic analysis if possible.   I'm not sure how feasible agnostic ASM is. :) If you're going to be super general from the start then `arg: Option<i64>` should be `args: Vec<AsmImmediate>`, which is overengineering.

pintc/src/pint_parser.lalrpop
otrho: Ahh, sorry, I made assumptions about `parse_int_immediate()` without refreshing myself on what it actually does.  I guess the function name should be `validate_int_immediate()`?

mohammadfawaz: Well... it does take a string and output an `Immediate` but the immediate could be a `B256` which we don't want here but do want in other places.

otrho (approved):

PR #1086, opened 2025-03-02
Initial support for dynamically sized arrays.

otrho:
This is a start in getting unsized arrays going.  It supports the passing unsized arrays as parameters to a predicate, but that's about it.  They can be used in union variants, but not in tuples or other arrays.  A new `__len()` intrinsic can be used to get their length.

There's a bunch of stuff which still needs to be done, but that's tracked in #1089.


mohammadfawaz (approved):
Good! Two comments: 
1. I think removing `Vector` and directly handling `UnsizedArray` in `lower_storage_accesses` would make more sense? Something I can do in my storage vector PR.
2. I suppose we plan on adding code to panic for OOB accesses? For vectors, I convert index accesses into the equivalent of: `((10 < __len(storage::v)? storage::v[10] : __panic_if(true));`. For morphisms, we won't need that because we're guaranteed in-bound accesses.

otrho:
> Good! Two comments:
> 
> 1. I think removing `Vector` and directly handling `UnsizedArray` in `lower_storage_accesses` would make more sense? Something I can do in my storage vector PR.

Cool.

> 2. I suppose we plan on adding code to panic for OOB accesses? For vectors, I convert index accesses into the equivalent of: `((10 < __len(storage::v)? storage::v[10] : __panic_if(true));`. For morphisms, we won't need that because we're guaranteed in-bound accesses.

I have added OOB checks in the ASM gen for `Index`.  You can see `PNCIF` instructions in the output, pretty much doing the above.

So perhaps a question is where is it better?  In a lowering pass or just in ASM gen?

Also, is it super wasteful to do a check for every single access?  As you say, an access inside a morphism using the provided index isn't needed. Or if there are multiple constant accesses, like `a[5] && a[10]` then the check need only be for the 10.  But optimising that would almost certainly require an IR I reckon.

PR #1098, opened 2025-03-06
Renaming `Vars` to `Args` in abi gen

### LINKED ISSUE #1096: Rename `Vars` from `abi-gen` to `Data` or `Args`
We seem to be referring to predicate data as "arguments" in the docs but the VM refers to them as "predicate data" (similar to call data I guess). 

In the JSON ABI we call `params` for parameters so `Args` might make more sense.

mohammadfawaz:
Closes #1096 

a few other small changes as well


pint-abi-gen/src/args.rs
```diff
@@ -1,41 +1,176 @@
...
-                    abi = Some(input.parse()?);
-                }
-                "contract" => {
-                    if contract.is_some() {
-                        return Err(input.error("duplicate `contract` argument"));
-                    }
-                    contract = Some(input.parse()?);
-                }
-                _ => return Err(input.error("unexpected argument")),
+//! Struct and impls for the predicate arguments `Args` type.
```
mohammadfawaz, line 1: This diff is weird.. I basically renamed the old `args.rs` to `macro_args.rs` and `vars.rs` to `args.rs`

pint-abi-types/src/lib.rs
```diff
@@ -1,27 +1,33 @@
...
     pub ty: TypeABI,
 }
 
+/// A storage variable
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct StorageVarABI {
+    pub name: String,
+    pub ty: TypeABI,
+}
+
```
mohammadfawaz, line 30: Even though these two types have the same fields, they refer to different things so I decided to split `ParamABI` into two separate types.

pint-abi-gen/src/args.rs
otrho: Did you use `git mv ...`?  Usually that works OK.

otrho (approved):

pint-abi-gen/src/args.rs
mohammadfawaz: Oh I did not. I'll try that next time

PR #1099, opened 2025-03-06
Add the ability to choose what test(s) to run in `validation_tests`

### LINKED ISSUE #1032: Add the ability to choose what test(s) to run in `validation_tests`
Would be nice to be able to run the following under `pint/tests`:
```console
cargo test -- --name simple_db.pnt
```
(or something similar).

lethalgem:
For `pint/tests` a filename can now be specified with

```sh
TEST_NAME=<filename.pnt> Cargo test
```
ex
```sh
TEST_NAME=unions_in_storage.pnt Cargo test
```

For `pintc/tests` a filename can now be specified with
```sh
TEST_NAME=<filename.pnt> Cargo test <directory>
```
ex
```sh
TEST_NAME=golf_array.pnt Cargo test arrays
```

The filename must be contained within the directory.

Closes #1032 

--------
Leaving the below for preservation:

I present to ya'll for discussion...

## Option A - Specifying tests via cargo test arguments
```sh
Cargo test <sub_dir or filename.pnt> -- [<sub_dir or filename.pnt>]
```
ex. To run all tests in the `arrays` directory:
```sh
cargo test arrays
```
ex. To run a specific test file `map_bad_type.pnt`:
```sh
cargo test map_bad_type.pnt
```
ex. To run a combination of all tests in the `arrays` directory and the specific test files `map_bad_type.pnt`, `const_folding_lets.pnt`
```sh
cargo test arrays -- map_bad_type.pnt const_folding_lets.pnt
```

Pros:
- Pretty
- Supports any combination of test directories and test names

Cons:
- We lose the heavy lifting of `libtest`. We're now using a custom test harness. So any feature we wanted from `libtest` will have to be manually implemented. There are some custom test harness crates out there and they all seem to agree that having feature parody with `libtest` is [near impossible](https://docs.rs/libtest-mimic/latest/libtest_mimic/#known-limitations-and-differences-to-the-official-test-harness). We aren't using any crate here, just our own minimal test harness.

## Option B - Specifying a test parameter
```sh
TEST_NAME=<filename.pnt> Cargo test <sub_dir>
```
ex. To run a specific test file `bad_const_array_init.pnt`:
```sh
TEST_NAME=bad_const_array_init.pnt Cargo test arrays
```

Pros:
- Keeps `libtest` functioning as normal. So we get to keep all of the normal arguments like `no-capture`, etc.

Cons:
- Ugly
- Currently doesn't support specifying multiple tests (which can be added easily), or a combination of directories and tests (kind of a PITA to add)

## My Opinion
I think Option A is the way to go. It gives the tester a _lot_ more control. And the custom test harness only applies to this directory. Using `libtest` with any other directory is trivial.

If we only ever need to specify a single test to run and not a combination of directories and multiple test names, then Option B is the way to go.


lethalgem:
After discussion with @mohammadfawaz, we're going to opt for Option B with the environment variable because we can't get rid of core features like `--no-capture`

mohammadfawaz (approved):
Great! Thanks.

PR #1100, opened 2025-03-07
Update to the latest Essential VM.

otrho:
In particular, breaking but welcome changes:
- STO, STOR and STOS now take the index as the last arg.
- RES returns the index in the stack to the reserved items.
- DROP introduced.


mohammadfawaz (approved):
Awesome, yeah much better. Thanks!

PR #1105, opened 2025-03-11
Do not preallocate memory for `KRNG` and `KREX`

mohammadfawaz:
This, therefore, removes the need for the `storage_expr_to_mem_idx` map in asm gen


pintc/src/asm_gen/asm_builder.rs
```diff
@@ -200,45 +168,66 @@ impl<'a> AsmBuilder<'a> {
...
+                        LODS,
+                        PUSH(1),
+                        ADD,
+                        LOD,
+                    ]);
                 }
 
+                // We can now free the memory used by KRNG OR KREX
+                asm.extend([PUSH(Self::KEY_RANGE_MEM_IDX_STACK_LOC), LODS, FREE]);
+
```
otrho, line 230: Do we need to do this?  I'm kinda thinking our model can be nodes just allocate as much as they want and we have a single `FREE` at the end.  I guess we could exhaust memory on a big loop which allocates?  But once we have the compute op everything will be made up of tiny nodes and so mid-execution `FREE` really won't be necessary IMO.

otrho (dismissed):
So it's possible for storage reads to happen inside a morphism body, yeah?  They can be anywhere now, not just the RHS of `let`.  So the scratch at stack[0] would be clobbered right now since they both use it.

But I'm planning to formalise the way we use stack[0] now, so this shouldn't be a problem, but yeah, I'll be changing this for storage reads too.

pintc/src/asm_gen/asm_builder.rs
mohammadfawaz: Yeah you're right.. no need to do this really. And memory for constraint nodes doesn't need to be clear at all.. it just gets dropped by the VM eventually anyways.

otrho (approved):

PR #1108, opened 2025-03-16
A few upgrades

mohammadfawaz:


pintc/tests/cli.rs
```diff
@@ -84,21 +84,21 @@ fn compile_errors() {
             .replace(input_file.path().to_str().unwrap(), "filepath"),
         expect_test::expect![[r#"
             Error: empty tuple types are not allowed
-               ╭─[filepath:1:19]
+               ╭─[ filepath:1:19 ]
```
mohammadfawaz, line 87: This seems to be a breaking change in `ariadne` even though they only released a patch release (`0.5.0` to `0.5.1`).

otrho (approved):

PR #1127, opened 2025-04-10
Consolidate `Type::Vector` and `Type::UnsizedArray`

mohammadfawaz:


mohammadfawaz:
There is more work related to vectors that needs to be done. This just cleans things up a bit and re-enable vectors so we can proceed.

otrho (approved):

